PR #16, opened 2021-05-05
Support file extension with dot

### LINKED ISSUE #15: File extensions cannot include dots
Currently we assume that everything after the last dot is the extension.

This has the consequence that TAR files are supported:

```python
backend.path('my.tar, '1.0.0')
```
```
/host/my/1.0.0/my-1.0.0.tar
```

but TAR.GZ files not:

```python
backend.path('my.tar.gz', '1.0.0')
```
```
/host/my.tar/1.0.0/my.tar-1.0.0.gz
```

Of course, it's not straightforward to solve, since we need to split of the extension to insert the version number. Therefore I suggest to add an optional `ext` argument that allows to specify the extension, i.e.:

```python
backend.path('my.tar.gz', '1.0.0', ext='tar.gz')
```
```
/host/my/1.0.0/my-1.0.0.tar.gz
```

frankenjoe:
Closes #15 

### Example

```python
backend.path('my.tar.gz', '1.0.0', ext='tar.gz')
```
```
/host/my/1.0.0/my-1.0.0.tar.gz
```


frankenjoe:
When I run the tests locally I see a lot of:

```
E           RuntimeError: {
E             "errors" : [ {
E               "status" : 401,
E               "message" : "Unauthorized"
E             } ]
E           } 401
```

guess I still need an account for `https://audeering.jfrog.io/artifactory`.

audbackend/core/backend.py
```diff
@@ -272,7 +298,17 @@ def path(
...
+            name, ext = os.path.splitext(file)
+        else:
+            if not ext.startswith('.'):
+                ext = '.' + ext
+            if not path.endswith(ext):
+                raise ValueError(
+                    f"Invalid path name '{path}', "
+                    f"does not end on '{ext}'."
+                )
+            name = file[:-len(ext)]
```
hagenw, line 311: Before we implement it here, should we maybe add it to [audeer.file_extension()](https://audeering.github.io/audeer/api-audeer.html#file-extension) and reuse it here?

hagenw: But I just realized that it doesn't make much sense to add `ext=` to `audeer.file_extension()`. If you know the extension, why should you use that function. We could add an argument where you specify the number of dots in your file extension, but I'm also not much in favor of it. So for now I would propose to just stick with your implementation here.

PR #17, opened 2021-05-06
Support empty file extension

frankenjoe:
I realized we were not supporting the case that the file extension is empty. Now the following works, too:

### Example

```python
backend.path('my.file.without.ext', '1.0.0', ext='')
```
```
/host/my.file.without.ext/1.0.0/my.file.without.ext-1.0.0
```


hagenw:
What are you planning with the backend :p

But yeah, changes make totally sense.

Issue #10, opened 2021-03-26
Replace _glob() with a simpler function


hagenw:
Currently, a `Backend` class needs to implement a `_glob()` function, which according docstring has to be able to scan into sub-directories and support wildcards. Implementing it for `Artifactory` and `FileSystem` was quite simple, but for other backends we might implement in the future it might become tricky. So I wonder if we should replace it with a simpler function that only returns the file names of a directory (and maybe another one that returns sub-directories). If we want to support recursive search and wildcards, we should implement this in `Backend.glob()` based on this simpler function(s).

hagenw:
A first step to replace `glob()` would be to add a method that allows to list the content of a folder on the backend. 

I'm not sure if this will speed `audb.available()` up for Artifactory as we would have to first get a list of all databases and then for each database a list of all versions. But at least we could try as the current solution using `glob()` is also not very fast:

```python
>>> timeit.timeit("df = audb.available()", number=1, setup="import audb")
60.44373281300068
```

frankenjoe:
> A first step to replace `glob()` would be to add a method that allows to list the content of a folder on the backend.

Sounds like a plan :)

frankenjoe:
I'd like to pick up this topic again. Reason: I have implemented a backend for [Google Cloud Storage](https://cloud.google.com/storage), which was straight-forward except for the `glob` function. We could implement it [like this](https://gist.github.com/agraubert/e3a5c4e4897aacbed82559453da128a8), but I think on the long run it's better if we don't force users to implement a complex backend function like `glob`.

The question is how we should proceed now. I see two options:

1. deprecate and remove at some point, new backends are not required to implement it
2. Implement glob in `Backend` using some simpler functions provided by the derived classes (e.g. this is the implementation in [cpython](https://github.com/python/cpython/blob/main/Lib/glob.py)

[...]

PR #18, opened 2021-05-10
Glob spcific folder

frankenjoe:
Add `folder` argument to `Backend.glob()` to restrict search to a specific folder on the backend.

### DIFF
--- a/audbackend/core/backend.py
+++ b/audbackend/core/backend.py
@@ -176,6 +176,7 @@ class Backend:
     def _glob(
             self,
             pattern: str,
+            path: str,
     ) -> typing.List[str]:  # pragma: no cover
         r"""Return matching files names."""
         raise NotImplementedError()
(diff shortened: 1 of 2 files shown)


hagenw:
Would it make sense to tackle directly https://github.com/audeering/audbackend/issues/10 and replacing `glob()` by a function that lists the content of a given folder instead?

Maybe, it is safer if we stay with `glob()` for a moment and add another function that can list the content of a folder, so I guess your changes are still worth applying them.

hagenw:
Otherwise all changes are fine and this would be ready to merge.

frankenjoe:
> Maybe, it is safer if we stay with `glob()` for a moment and add another function that can list the content of a folder, so I guess your changes are still worth applying them.

Yes, I think we should stay with it until we have the time to properly replace it.

PR #19, opened 2021-06-03
Split up backend.py into several files

hagenw:
To make the backend implementation easier to understand, I propose to split up the `backend.py` file into:
* `api.py`
* `artifactory.py`
* `backend.py`
* `filesystem.py`


frankenjoe:
Thanks, your changes make sense.

PR #20, opened 2021-06-17
DOC: link to audfactory for authentication

hagenw:
This changes the authentication documentation by providing a link to the `audfactory` documentation as single source of truth.

### DIFF
--- a/docs/authentication.rst
+++ b/docs/authentication.rst
@@ -22,7 +22,7 @@ Alternatively, export the credentials as environment variables:
     export ARTIFACTORY_API_KEY="MY_API_KEY"
 
 The environment variables will be applied to all servers.
-You might loose access to artifacts on servers
+You might lose access to artifacts on servers
 that are setup for anonymous access
 as it will always try to authenticate
 with the given username and password.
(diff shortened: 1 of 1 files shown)


hagenw:
I just realized that we have the problem of not having one single source of truth for this documentation as we also have it at https://audeering.github.io/audfactory/usage.html#authentication

Should I maybe update the documentation there and add simply a link to there on this page?

docs/authentication.rst
```diff
@@ -1,17 +1,30 @@
...
+
+Alternatively, export the credentials as environment variables:
 
 .. code-block:: bash
 
     export ARTIFACTORY_USERNAME="MY_USERNAME"
     export ARTIFACTORY_API_KEY="MY_API_KEY"
+
+The environment variables will be applied to all servers.
+You might loose access to artifacts on servers
```
frankenjoe, line 25:
```suggestion
You might lose access to artifacts on servers
```

frankenjoe:
> Should I maybe update the documentation there and add simply a link to there on this page?

yes, I think that's the best solution

hagenw:
OK, I created https://github.com/audeering/audfactory/pull/24 and will update this pull request as soon as the new docs of `audfactory` are online.

hagenw:
The new page now looks like this (was not able to include the picture in the description above)

image

PR #21, opened 2021-07-13
Add audbackend.Repository

hagenw:
This adds the `audbackend.Repository` class, which we can then easily reuse in every project that needs to address repositories.
The users there don't have to call the actual `audbackend` package, but we could directly import in the `__ini__.py` file. E.g. for `audb` it would be

```python
from audbackend import Repository
```

instead of

```python
from audb.core.repository import Repository
```

image


frankenjoe:
Cool, that is indeed usefui.

PR #23, opened 2021-07-15
FIX: ignore empty strings in backend.join()

### LINKED ISSUE #22: Ignore empty strings in backend.join()
Currently, `backend.join()` does not ignore empty strings:

```python
backend.join('', '', 'file.txt')
```
```
//file.txt
```

This can lead to malformed URLs, e.g.:

```python
backend.path(path, '1.0.0')
```
```
http://host/repository///file/1.0.0/file-1.0.0.txt
```

frankenjoe:
Closes #22

Fixes the problem and adds a test for it.

### Example:

```python
backend = audbackend.create('artifactory', 'http://host', 'repository')
backend.join('', '', 'file.txt')
```
```
file.txt
```

### DIFF
--- a/audbackend/core/backend.py
+++ b/audbackend/core/backend.py
@@ -213,7 +213,9 @@ class Backend:
             path joined by :attr:`Backend.sep`
 
         """
-        return self.sep.join([path] + [p for p in paths])
+        paths = [path] + [p for p in paths]
+        paths = [path for path in paths if path]
+        return self.sep.join(paths)
 
     def latest_version(
             self,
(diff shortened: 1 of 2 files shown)


audbackend/core/backend.py
```diff
@@ -213,7 +213,9 @@ def join(
             path joined by :attr:`Backend.sep`
 
         """
-        return self.sep.join([path] + [p for p in paths])
+        paths = [path] + [p for p in paths]
+        paths = [path for path in paths if path]
```
hagenw, line 217: Could you add a comment above this line stating that it removes `''` path entries. Because otherwise it might be hard to get later on.

frankenjoe: done

PR #27, opened 2021-09-28
Catch 403 Error for Artifactory backend paths

### LINKED ISSUE #26: Artifactory backend migth fail with 403 error
When using `Artifactory._exists()` or `Artifactory._glob()` it might fail with
```
requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: ...
```
as we only cover `RuntimeError` when catching errors from Artifactory, but with the newest release from `dohq-artifactory` it seems that we can also get a `requests.exceptions.HTTPError`

hagenw:
Closes #26 

When accessing a non-existing path or a path where you don't have the permission to access it, you might experience a 403 error returned from Artifactory. This started happening only since a few days and most likely it depends on the version of `dohq-artifactory` and maybe if you have anonymous access enabled or not.

Here, we don't care, but just catch the corresponding error in order to return a proper `False` for `exists()` and an empty list for `glob()`.

### DIFF
--- a/audbackend/core/artifactory.py
+++ b/audbackend/core/artifactory.py
@@ -1,3 +1,4 @@
+import requests
 import typing
 
 import audfactory
@@ -53,9 +54,12 @@ class Artifactory(Backend):
     ) -> bool:
         r"""Check if file exists on backend."""
         try:
-            # Can lead to RuntimeError: 404 page not found
+            # Can lead to
+            # RuntimeError: 404 page not found
+            # or
+            # requests.exceptions.HTTPError: 403 Client Error
             return audfactory.path(path).exists()
-        except RuntimeError:  # pragma: nocover
+        except self._non_existing_path_error:  # pragma: nocover
             return False
 
     def _get_file(
@@ -84,7 +88,7 @@ class Artifactory(Backend):
         path = audfactory.path(url)
         try:
             result = [str(x) for x in path.glob(pattern)]
-        except RuntimeError:  # pragma: no cover
+        except self._non_existing_path_error:  # pragma: nocover
             result = []
         return result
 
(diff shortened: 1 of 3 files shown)


hagenw:
I was not really able to create a test to catch the new 403 error as this would require setting up another repository and most likely disable anonymous access, etc.
So for the short term I would propose we just include the fix, but stay with the non-coverage testing.

audbackend/core/artifactory.py
```diff
@@ -123,3 +127,18 @@ def _versions(
...
+    r"""Error expected for non-existing paths.
+
+    If a user has no permission to a given path
+    or the path does not exists :func:`audfactory.path`
+    might return a
+    ``RuntimeError: 404 page not found``
+    or
+    ``requests.exceptions.HTTPError: 403 Client Error``
+    error,
+    which might depend on the instaleld ``dohq-artifactory``
```
frankenjoe, line 141:
```suggestion
    which might depend on the installed ``dohq-artifactory``
```

audbackend/core/artifactory.py
```diff
@@ -53,9 +54,12 @@ def _exists(
...
         r"""Check if file exists on backend."""
         try:
-            # Can lead to RuntimeError: 404 page not found
+            # Can lead to
+            # RuntimeError: 404 page not found
+            # or
+            # requests.exceptions.HTTPError: 403 Client Error
             return audfactory.path(path).exists()
-        except RuntimeError:  # pragma: nocover
+        except self._non_existing_path_error:  # pragma: nocover
```
frankenjoe, line 62: Could it make sense to catch any `Exception` here to avoid the code breaks again in the future?

audbackend/core/artifactory.py
hagenw: Sorry, too late.

As we need this fix as fast as possible I decided to go without review.

audbackend/core/artifactory.py
hagenw: In general it is not a good idea to catch simply everything, because you might also get an exception because you use your keyboard to interrupt or you don't have an internet connection. Those cases we would then also catch.

On the other hand it will make sense easier for us to maintain.

PR #31, opened 2022-02-24
Check for path name before creating archive

hagenw:
We check that a given path name includes only allowed chars.
But for `audbackend.Backend.put_archive()` we did first create the archive and only afterwards checked for the correct path name.
This has two disadvantages:
* it can take some time before the error is raised as the archive is first created
* on Windows it might fail before as it cannot create the archive locally, compare https://github.com/audeering/audb/runs/5320867038?check_suite_focus=true

I added a test for using a non-supported archive name, which fails for the `master` under Widnows with an `OSError`, see https://github.com/audeering/audbackend/runs/5321158317?check_suite_focus=true
I then added the check for the correct name at the beginning of `put_archive()`, which now raises our custom `ValueError`.

### DIFF
--- a/audbackend/core/backend.py
+++ b/audbackend/core/backend.py
@@ -381,7 +381,7 @@ class Backend:
             FileNotFoundError: if one or more files do not exist
 
         """
-        utils.check_path_for_allowed_chars(dst_path)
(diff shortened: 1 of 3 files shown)


hagenw:
> We need to merge #32 first before we continue here.

Done

Issue #37, opened 2022-04-27
ls() does not raise expected error


frankenjoe:
```python
import audbackend


backend = audbackend.create(
    'file-system',
    'host',
    'repo',
)
backend.ls('does-not-exist')
```
```
FileNotFoundError: [Errno 2] No such file or directory: '.../host/repo/does-not-exist'
```

and

```python
backend = audbackend.create(
    'artifactory',
    'https://artifactory.audeering.com/artifactory',
    'unittests-public-local',
)
backend.ls('does-not-exist')
```
```
FileNotFoundError: [Errno 2] No such file or directory: '.../unittests-public-local/does-not-exist'
```

But according to the [documentation](https://audeering.github.io/audbackend/api.html#audbackend.Backend.ls) it should raise a `RuntimeError`. However, in all other functions we raise a `FileNotFoundError`, so I would call it a bug and recommend we change the API accordingly.

But there is another problem: currently we let the child implementation raise the error, i.e. we do not have control what error it will actually raise. A proper solution would be to first check if the `path` exists and then call the child class. But here we run into another problem since [exists()](https://audeering.github.io/audbackend/api.html#audbackend.Backend.exists)  only works for files. Instead we could directly call `_exists()` on the child class, which should also work for directories, but the problem is that we need to convert it to backend path first, but [path()](https://audeering.github.io/audbackend/api.html#audbackend.Backend.path) also works only for files.

frankenjoe:
I guess we need to change at least the implementation of `path()` to also accept folders or introduce another function that converts any file or folder to the backend path.

frankenjoe:
I checked what happens if we currently provide `version=None` to `path()` and for `FileSystem` it works:

```python
backend = audbackend.create(
    'file-system',
    'host',
    'repo',
)

root = 'test'
backend.path(root, None)
```
```
.../host/repo/test
```

But not for `Artifactory`:

backend = audbackend.create(
    'artifactory',
    'https://artifactory.audeering.com/artifactory',
    'unittests-public-local',
)

root = 'com'
backend.path(root, None)
```
```
.../unittests-public-local//com/com-None
```

So one option would be that we officially support `version=None` with `path()` and fix it for `Artifactory'.

**Comment by frankenjoe** [comment]
Mhh, also `FileSystem` does not always work as expected with `version=None`:

```python
backend = audbackend.create(
    'file-system',
    'host',
    'repo',
)

root = 'test/1.0.0'
backend.path(root, None)
```
```
[...]

PR #38, opened 2022-04-27
WIP: Add Backend.folder()

frankenjoe:
As discussed in #37 we need a method `folder()` that converts a folder path to its backend form. It is now used within `ls()` to check if the path exists and make sure we always raise a `FileNotFoundError`error if it does not. It also fixes the docstring that was wrongly stating we raise a `RuntimeError` in that case. A test is added to verify we raise the correct error.

image


frankenjoe:
#39 implements an alternative solution without introducing a new API function.

frankenjoe:
Closing in favor of #46

PR #39, opened 2022-04-27
WIP: Force same file structure on all backends

frankenjoe:
This is alternative solution for #38, but also introduces a more fundamental change regarding the file structure we use on the backends.

Instead of adding a new API function `folder()`, this changes the implementation of `_path()` and as a side effect it can now cope with folders, too. It's probably more elegant as we do not introduce another abstract function and now `_path()` is also easier to implement. I also don't see why a user would need a `folder()` function.

The core change we propose here is to create the file structure already in `path()` and let `_path()` only add host, repository and possibly change the file separator. I think this makes a lot of more sense, since now we ensure that version `1.0.0` of `sub/dir/file.txt` will be stored as `sub/dir/file/1.0.0/file-1.0.0.txt` on all backends. So far this was left to the child implementation, e.g. the backend was allowed to store the file as `sub/dir/file_1.0.0.txt`. But I don't think this is good design. 1. we assume that the developer chooses a format that properly encodes the version, 2. we have the unwanted side-effect that `ls()` may return different results depending on the backend.

### DIFF
--- a/audbackend/core/artifactory.py
+++ b/audbackend/core/artifactory.py
@@ -31,23 +31,6 @@ class Artifactory(Backend):
         r"""MD5 checksum of file on backend."""
         return audfactory.checksum(path)
 
-    def _path(
-            self,
-            folder: str,
-            name: str,
-            ext: str,
-            version: str,
-    ) -> str:
-        r"""File path on backend."""
-        server_url = audfactory.url(
-            self.host,
-            repository=self.repository,
-            group_id=audfactory.path_to_group_id(folder),
-            name=name,
-            version=version,
-        )
-        return f'{server_url}/{name}-{version}{ext}'
-
     def _exists(
             self,
             path: str,
@@ -104,6 +87,18 @@ class Artifactory(Backend):
             group_id=audfactory.path_to_group_id(path),
         )
         return [p.name for p in audfactory.path(path)]
+
+    def _path(
+            self,
+            path: str,
+    ) -> str:
+        r"""File path on backend."""
+        url = audfactory.url(
+            self.host,
+            repository=self.repository,
+        )
+        url = f'{url}/{path}'
+        return url
 
     def _put_file(
             self,
--- a/audbackend/core/backend.py
+++ b/audbackend/core/backend.py
@@ -284,10 +284,7 @@ class Backend:
 
     def _path(
             self,
-            folder: str,
-            name: str,
-            ext: str,
-            version: str,
+            path: str,
     ) -> str:  # pragma: no cover
         r"""File path on backend."""
         raise NotImplementedError()
@@ -334,6 +331,7 @@ class Backend:
         """
         utils.check_path_for_allowed_chars(path)
         folder, file = self.split(path)
+
         if ext is None:
             name, ext = os.path.splitext(file)
         elif ext == '':
@@ -347,7 +345,15 @@ class Backend:
                     f"does not end on '{ext}'."
                 )
             name = file[:-len(ext)]
-        return self._path(folder, name, ext, version)
+
+        path = self.join(
+            folder,
+            name,
+            version,
+            f'{name}-{version}{ext}',
+        )
+
+        return self._path(path)
 
     def put_archive(
             self,
--- a/audbackend/core/filesystem.py
+++ b/audbackend/core/filesystem.py
@@ -33,28 +33,6 @@ class FileSystem(Backend):
         r"""MD5 checksum of file on backend."""
         return utils.md5(path)
 
-    def _path(
-            self,
-            folder: str,
-            name: str,
-            ext: str,
-            version: str,
-    ) -> str:
-        r"""File path on backend."""
-        path = os.path.join(
-            self.host,
-            self.repository,
-            folder.replace(self.sep, os.path.sep),
-            name,
-        )
-        if version is not None:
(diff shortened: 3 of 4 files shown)


hagenw:
There is one problem with your proposed changes, they assume that every backend stores files with `sub/dir/file/1.0.0/file-1.0.0.txt`, but this might not be the case. E.g. a backend might not encode the version of a file as a folder and within the file name, but stores it as metadata in a database and it needs the version to be handed over as argument. In principle a user can still implement it inside `_path` by re-engineering `folder`, `name`, `version` from `path`, but why not stick with it in the first place?

> Instead of adding a new API function folder(), this changes the implementation of _path() and as a side effect it can now cope with folders, too.

This I do not understand, can you try to provide an example for this.

frankenjoe:
> There is one problem with your proposed changes, they assume that every backend stores files with sub/dir/file/1.0.0/file-1.0.0.txt, but this might not be the case.

Exactly, currently it is left to the user under which path a file is stored. But it also means that we cannot know what `ls()` returns. For example if `backend1` stores version `1.0.0` of `sub/dir/file.txt` as `sub/dir/file/1.0.0/file-1.0.0.txt`, then `ls('sub/dir/file/')` returns a list of versions. But if `backend2` stores the file as `sub/dir/file-1.0.0.txt` it returns an empty list. That means that the use of `ls()` is not safe, as the result depends on the platform.

>  In principle a user can still implement it inside _path by re-engineering folder, name, version from path, but why not stick with it in the first place?

Yes, in principle that is possible of course :)

> This I do not understand, can you try to provide an example for this.

Currently, `_path()` does not work with folders as it expects path, version and extension and the latter do not have a meaning for folders. Now, only the path is passed and it doesn't matter if the argument is a folder or a file, since the function only has to add host and repository to it and possibly change the delimiter.

hagenw:
> That means that the use of ls() is not safe, as the result depends on the backend.

This sounds to me like we have not implemented `ls()` in a good way. I think it should never return any versions, also not as part of the file names. Because if you look at `backend.exists(path, version)` we provide `path` also without version, and I would assume we then also need to have `backend.ls(path, version)` instead of `backend.ls(path)`

hagenw:
E.g. if we have

```python
backend.path('media/archive1.zip', '1.0.0')
```
Then I would expect
```python
>>> backend.ls('media', '1.0.0')
'media/archive1.zip'
```

frankenjoe:
> I would assume we then also need to have `backend.ls(path, version)` instead of `backend.ls(path)`

But means we can no longer use it to search through the folder structure of the backend. E.g. other than in `audb` we currently  models / benchmarks with different group IDs so you have (simplified) `sub/dir/model.zip` and `sub/model.zip`. To list the models we currently do recursive search through all folders with or `glob('sub/**/model.zip')`. In principle we could change that and implement `ls()` in a way that i automatically does a recursive search and only return the files that match at the leaves but not any other content.

hagenw:
> > I would assume we then also need to have `backend.ls(path, version)` instead of `backend.ls(path)`
> 
> But means we can no longer use it to search through the folder structure of the backend. E.g. other than in `audb` we currently models / benchmarks with different group IDs so you have (simplified) `sub/dir/model.zip` and `sub/model.zip`. To list the models we currently do recursive search through all folders with or `glob('sub/**/model.zip')`. In principle we could change that and implement `ls()` in a way that i automatically does a recursive search and only return the files that match at the leaves but not any other content.

Yes, let's assume we have:

```python
backend.path('media/archive1.zip', '1.0.0')
backend.path('media/sub/archive2.zip', '1.0.0')
```

Then we could do:

```python
>>> backend.ls('media', '1.0.0')
'media/archive1.zip'
>>> backend.ls('media', '1.0.0', recursive=True)
'media/archive1.zip'
'media/sub/archive2.zip'
```

frankenjoe:
Yes, makes sense. Let's pursuit this as another solution towards #37 and #10. Will put WIP to this PR.

frankenjoe:
> Yes, let's assume we have:
> 
> ```python
> backend.path('media/archive1.zip', '1.0.0')
> backend.path('media/sub/archive2.zip', '1.0.0')
> ```
> 
> Then we could do:
> 
> ```python
> >>> backend.ls('media', '1.0.0')
> 'media/archive1.zip'
> >>> backend.ls('media', '1.0.0', recursive=True)
> 'media/archive1.zip'
> 'media/sub/archive2.zip'
> ```

Actually this is also problematic. Cause as far as I see you propose we now return the path as it is represented to the user, e.g. not `'media/archive/1..0/archive-1.0.0.zip'` but `'media/archive1.zip'`. However, this might be problematic, since we currently only have `path()` to convert `'media/archive1.zip'` to `'media/archive/1..0/archive-1.0.0.zip'`, but not the other way round.

hagenw:
Yes, but I don't see another way to implement `backend.ls()`. What if the actual storage structure on the backend has nothing to do with a folder/file based structure in this case it makes no sense to return something different then the path as it is represented to the user.

frankenjoe:
Right, the current implementation which returns the backend path and not the virtual path does not make sense at all. That means we will have to extend the interface with a method `_inverse_path()` or something like this.

frankenjoe:
Let's move the discussion to #40

frankenjoe:
Closing in favor of #46

Issue #46, opened 2022-12-05
Fix audbackend.Backend interface


frankenjoe:
As discussed in https://github.com/audeering/audbackend/issues/40 there are two problems with the current backend interface.

1. The method `Backend.path()` assumes there is a deterministic way to convert a virtual path to a backend path. This is probably too restrictive. E.g. think of a backend which assigns a random ID to a file when it is uploaded.
2. The method `Backend.ls()` takes as input a folder in the backend format. But this assumes that the concept of a folder exists for every backend, which is again too restrictive. E.g. think of a backend where files are stored in a database under some key.

To overcome these issues I propose we do the following:

- [x] Remove the method `Backend.path()` and the according abstract method `Backend._path()`. Instead we forward the virtual path to all abstract functions. ~~For backward compatibility we implement `FileSystem.path()` and `Artifactory.path()` directly in the child class.~~. See https://github.com/audeering/audbackend/pull/47
- [x] Fix `Backend.ls()` that it accepts a folder in virtual format. ~~The current implementation is a bug, but maybe we still want to implement `FileSystem.ls()` and `Artifactory.ls()` for backward compatibility and choose a new method name.~~, see https://github.com/audeering/audbackend/pull/53
- [x] Remove `Backend.glob()`, see https://github.com/audeering/audbackend/pull/79
- [x] Implement a backend without folders and randomly generated file names to verify the interface supports such a backend now

frankenjoe:
@hagenw I might dedicate some time to work on this, but wanted to get your feedback first if you agree the suggested approach makes sense.

hagenw:
Regarding `Backend.path()` I slightly disagree. I think we need a deterministic way to get a path, otherwise we will not be able to find a file again on the backend. Converting the virtual path can be done by a random process, but then we need to have a mapping/database that stores the connection between the virtual path and the actual path on the backend.
But it might be that a user would never need `Backend.path()`, then we could indeed remove it. But then I would also remove `FileSystem.path()` and `Artifactory.path()`.

I would indeed change the behavior of `Backend.ls()`, and think it should also be fine to add breaking changes to `FileSystem.ls()` and `Artifactory.ls()` and update packages where we use this code to the new behavior.

frankenjoe:
> Converting the virtual path can be done by a random process, but then we need to have a mapping/database that stores the connection between the virtual path and the actual path on the backend.

Sure, but that lies in the responsibility of the backend implementation.

[...]

PR #47, opened 2022-12-05
WIP: remove Backend.path()

frankenjoe:
Relates to #46

### DIFF
--- a/audbackend/core/artifactory.py
+++ b/audbackend/core/artifactory.py
@@ -36,41 +36,6 @@ class Artifactory(Backend):
         path = self._path(path, version, ext)
         return audfactory.checksum(path)
 
-    def _path(
-            self,
-            path: str,
-            version: str,
-            ext: str,
-    ) -> str:
-        r"""Convert to backend path.
-
-        Format: <server>/<folder>/<version>/<basename>-<version>.<ext>
-
-        """
-        utils.check_path_for_allowed_chars(path)
-
-        folder, file = self.split(path)
-
-        if ext is None:
-            name, ext = os.path.splitext(file)
-        elif ext == '':
-            name = file
-        else:
-            if not ext.startswith('.'):
-                ext = '.' + ext
-            name = file[:-len(ext)]
-
-        utils.check_path_ends_on_ext(path, ext)
-
-        server_url = audfactory.url(
-            self.host,
-            repository=self.repository,
-            group_id=audfactory.path_to_group_id(folder),
-            name=name,
-            version=version,
-        )
-        return f'{server_url}/{name}-{version}{ext}'
-
     def _exists(
             self,
             path: str,
@@ -134,6 +99,45 @@ class Artifactory(Backend):
         )
         return [p.name for p in audfactory.path(path)]
 
+    def _path(
+            self,
+            path: str,
+            version: typing.Optional[str],
+            ext: str,
+    ) -> str:
+        r"""Convert to backend path.
+
+        Format: <server>/<folder>/<version>/<basename>-<version>.<ext>
+
+        """
+        utils.check_path_for_allowed_chars(path)
+
+        folder, file = self.split(path)
+
+        if ext is None:
+            name, ext = os.path.splitext(file)
+        elif ext == '':
+            name = file
+        else:
+            if not ext.startswith('.'):
+                ext = '.' + ext
+            name = file[:-len(ext)]
+
+        utils.check_path_ends_on_ext(path, ext)
+
+        path = audfactory.url(
+            self.host,
+            repository=self.repository,
+            group_id=audfactory.path_to_group_id(folder),
+            name=name,
+            version=version,
+        )
+
+        if version is not None:
+            path = f'{path}/{name}-{version}{ext}'
+
+        return path
+
     def _put_file(
             self,
             src_path: str,
@@ -162,14 +166,12 @@ class Artifactory(Backend):
             ext: str,
     ) -> typing.List[str]:
         r"""Versions of a file."""
-        root, name = self.split(path)
-        group_id = audfactory.path_to_group_id(root)
-        vs = audfactory.versions(
-            self.host,
-            self.repository,
-            group_id,
-            name,
-        )
+        path = self._path(path, None, ext)
+        path = audfactory.path(path)
+        try:
+            vs = [os.path.basename(str(p)) for p in path if p.is_dir]
+        except (FileNotFoundError, RuntimeError):
+            vs = []
         return vs
 
     _non_existing_path_error = (RuntimeError, requests.exceptions.HTTPError)
@@ -182,7 +184,7 @@ class Artifactory(Backend):
     or
     ``requests.exceptions.HTTPError: 403 Client Error``
     error,
-    which might depend on the instaleld ``dohq-artifactory``
+    which might depend on the installed ``dohq-artifactory``
     version. So we better catch both of them.
 
     """
--- a/audbackend/core/artifactory.py
+++ b/audbackend/core/artifactory.py
@@ -1,15 +1,17 @@
+import os
 import requests
 import typing
 
 import audfactory
 
+from audbackend.core import utils
 from audbackend.core.backend import Backend
 
 
 class Artifactory(Backend):
     r"""Artifactory backend.
 
-    Stores files and archives on Artifactory.
+    Store files and archives on Artifactory.
 
     Args:
         host: host address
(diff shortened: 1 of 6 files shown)


audbackend/core/utils.py
```diff
@@ -9,6 +11,40 @@
 BACKEND_ALLOWED_CHARS_COMPILED = re.compile(BACKEND_ALLOWED_CHARS)
 
 
+def check_path_and_ext(
```
frankenjoe, line 14: Instead of converting the virtual backend path directly, we only call this function in the interface. It applies a couple of small checks, see list below.

tests/test_backend.py
```diff
@@ -149,9 +150,6 @@ def test_archive(tmpdir, files, name, folder, version, tmp_root, backend):
         tmp_root=tmp_root,
     ) == files_as_list
 
-    with pytest.raises(ValueError):
```
frankenjoe, line 152: This is already covered by the new tests below.

tests/test_backend.py
```diff
@@ -578,12 +545,24 @@ def test_versions(tmpdir, file_name, ext, backend):
...
     backend.put_file(local_file, remote_file, '1.0.0', ext=ext)
     assert backend.versions(remote_file, ext=ext) == ['1.0.0']
     assert backend.latest_version(remote_file, ext=ext) == '1.0.0'
+
+    # v2
     backend.put_file(local_file, remote_file, '2.0.0', ext=ext)
     assert backend.versions(remote_file, ext=ext) == ['1.0.0', '2.0.0']
     assert backend.latest_version(remote_file, ext=ext) == '2.0.0'
+
+    # v3 with a different extension
```
frankenjoe, line 563: This tests that versions of files with same path but a different extension are not returned.

audbackend/core/backend.py
```diff
@@ -286,73 +308,6 @@ def ls(
...
 
-    def _path(
-            self,
-            folder: str,
-            name: str,
-            ext: str,
-            version: str,
-    ) -> str:  # pragma: no cover
-        r"""File path on backend."""
-        raise NotImplementedError()
```
hagenw, line 297: In the single backend you still implement `_path()`.
But I guess with removing it from `Backend` you indicate that a developer don't has to implement it, correct.
Which means inside `Artifactory` and `Filesystem` you just make it out of convenience?

audbackend/core/backend.py
```diff
@@ -286,73 +308,6 @@ def ls(
         """
         return sorted(self._ls(path))
```
hagenw, line 309: For `ls()` we are missing a check if the path contains any non allowed chars. I guess you do not check here as it otherwise does not need `utils.check_path_and_ext()`

audbackend/core/backend.py
```diff
@@ -530,12 +491,13 @@ def split(
         """
```
hagenw, line 491: `split()` is also missing a check for invalid chars in `path`

audbackend/core/backend.py
```diff
@@ -398,23 +356,21 @@ def put_archive(
...
-                raise FileNotFoundError(
-                    errno.ENOENT, os.strerror(errno.ENOENT), path,
-                )
+                utils.raise_file_not_found_error(path)
 
         with tempfile.TemporaryDirectory(dir=tmp_root) as tmp:
-            _, archive_name = self.split(dst_path)
-            archive = audeer.path(tmp, f'{archive_name}-{version}.zip')
+
+            archive = audeer.path(tmp, os.path.basename(dst_path))
```
hagenw, line 363: We need to update `setup.cfg` to require version `1.18.0` as we added `audeer.path()` there.

tests/test_backend.py
```diff
@@ -202,6 +200,19 @@ def test_errors(tmpdir, backend):
         file_name,
     )
 
+    with pytest.raises(ValueError):  # extension does not match path
```
hagenw, line 203: As you test with `pytest.raises` already could you add a check for the correct error message with `match`, and also for the other tests.

audbackend/core/backend.py
```diff
@@ -455,27 +413,30 @@ def put_file(
...
+        dst_path, ext = utils.check_path_and_ext(dst_path, ext)
         if not os.path.exists(src_path):
-            raise FileNotFoundError(
-                errno.ENOENT, os.strerror(errno.ENOENT), src_path,
-            )
+            utils.raise_file_not_found_error(src_path)
 
-        dst_path = self.path(dst_path, version, ext=ext)
-
-        # skip if file with same checksum exists on backend
```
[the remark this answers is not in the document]
frankenjoe: So far we returned the real path on the backend and not the virtual one. Returning the virtual path means we basically return the value of the `dst_path` argument, which makes not so much sense I guess. Also, if we return the virual path here, it means there is no (official) way for the user to ever access the real path on the backend, but maybe that's ok. If we decide to not return the real path on the backend, I would return nothing.

tests/test_backend.py
frankenjoe: yes

audbackend/core/backend.py
frankenjoe: yes

audbackend/core/backend.py
frankenjoe: Here I was not sure if we really need the check since we are actually not accessing the backend. But if you think we should, we can add such a check of course.

tests/test_backend.py
frankenjoe: done

audbackend/core/backend.py
frankenjoe: done

tests/test_backend.py
hagenw: Interesting, seems we are lucky that this has had no unwanted consequences in other packages.

audbackend/core/backend.py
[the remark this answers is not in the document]
hagenw: From a user perspective, I find it easier if we have only a single path concept, which means we restrict us to the virtual paths. So updating the code to return nothing seems make the most sense to me.

As we also have `remove_file()` there is not really a need for a user to know the exact path. It only makes sense if you are the maintainer of the backend and need to know where files are located. But for now I would say it is fine to not provide a public method for that (e.g. a maintainer might use `audbackend.Artifactory._path()` to get the real path).

audbackend/core/backend.py
hagenw: You are right that it is not needed, but I think I would prefer to add it as we then have the same check whenever `path` is given. Such consistency is always easier to grasp for a user.

audbackend/core/backend.py
hagenw: I know that you didn't touched it. I just ask for the check to be consistent as with `split()`. 
I created https://github.com/audeering/audbackend/issues/48 so we don't need to change it here.

tests/test_backend.py
frankenjoe: Would you prefer that we fix it in a separate PR?

hagenw: Is it easy to separate from the rest? If not it should be fine to do it here.

audbackend/core/backend.py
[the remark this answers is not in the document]
frankenjoe: I moved the skipping part back into `put_file()`. The methods put_file()`, `put_archive()` and `remove_file()` now return `None`.

audbackend/core/backend.py
frankenjoe: Ok, we now check for invalid characters also in `Backend.split()` and `Backend.join()`.

audbackend/core/backend.py
```diff
@@ -455,26 +417,35 @@ def put_file(
...
-            utils.md5(src_path) == self._checksum(dst_path)
-        if not skip:
-            self._put_file(src_path, dst_path, verbose)
+        # skip if file with same checksum already exists
+        if self._exists(dst_path, version, ext):
+            checksum = self._checksum(dst_path, version, ext)
+            skip = utils.md5(src_path) == checksum
 
-        return dst_path
+        if not skip:
```
hagenw, line 435: Maybe the following is easier to understand?

```python
# skip if file with same checksum already exists
if not (
    self._exists(dst_path, version, ext)
    and utils.md5(src_path) == self._checksum(dst_path, version, ext)
):
```

Then you could also remove
```python
skip = False
```

frankenjoe: Yes, makes sense. Changed it accordingly.

audbackend/core/artifactory.py
```diff
@@ -105,30 +124,65 @@ def _ls(
...
+        folder = self._folder(path, ext)
+        folder = audfactory.path(folder)
+
+        try:
+            vs = [os.path.basename(str(f)) for f in folder if f.is_dir]
+        except (FileNotFoundError, RuntimeError):
+            vs = []
+
+        # filter out versions of files with different extension
+        vs = [v for v in vs if self._exists(path, v, ext)]
```
hagenw, line 183: I remember what the issue might be here. `audfactory.versions()` does not really handle this case as Artifactory was not designed to store the same file with different extensions (for whatever reasons).

As you now filter the versions here anyway could we stay with `audfactory.versions()` in the line before or was there another reason you replaced it?

hagenw: Good point, it should be relatively easy to deprecate `audfactory` as it is most likely only needed inside `audbackend`. The main thing we would need to add to `audbackend` would be the code of `audfactory.authentification()`.

So let's stay here without using `audfactory.versions()`.

hagenw: I created https://github.com/audeering/audbackend/issues/50

PR #52, opened 2022-12-07
Properly handle dots in backend folder on Artifactory

### LINKED ISSUE #49: Same virtual path can translate into different backend paths
As the following example shows the same virtual path can translate into different backend paths depending on the `ext` argument that we pass:

```python
backend = audbackend.create(
    'artifactory',
    'https://audeering.jfrog.io/artifactory',
    'unittests-public',
)

path = 'folder/file.foo.bar'

for ext in ['', '.foo', '.foo.bar']:
    backend_path = backend._path(
        path,
        '1.0.0',
        ext,
    )
    print(backend_path)
```
```
https://audeering.jfrog.io/artifactory/unittests-public/folder/file.foo.bar/1.0.0/file.foo.bar-1.0.0
https://audeering.jfrog.io/artifactory/unittests-public/folder/file.foo/1.0.0/file.foo-1.0.0.foo
https://audeering.jfrog.io/artifactory/unittests-public/folder/file/1.0.0/file-1.0.0.foo.bar
```

For consistency we have to enforce the same behavior for every backend implementation. So we should at least add a test for it.

### LINKED ISSUE #51: Different virtual paths can translate to the same backend path on Artifactory
As the following example shows using dots in the folders of a virtual backend path has the undesired side-effect that different virtual paths can translate to the same backend path on Artifactory

```python
for path in [
    'sub.folder/file.txt',
    'sub/folder/file.txt',
]:
    backend_path = backend._path(
        path,
        '1.0.0',
        '.txt',
    )
    print(backend_path)
```
```
https://audeering.jfrog.io/artifactory/unittests-public/sub/folder/file/1.0.0/file-1.0.0.txt
https://audeering.jfrog.io/artifactory/unittests-public/sub/folder/file/1.0.0/file-1.0.0.txt
```

If we use the file system as backend we get two different paths:

```
/host/repo/sub.folder/file/1.0.0/file-1.0.0.txt
/host/repo/sub/folder/file/1.0.0/file-1.0.0.txt
```

So either we fix the behavior for the Artifactory backend (if that is possible at all) or we have to forbid the use of dots in the folders of a virtual path.

frankenjoe:
Closes #51  

Fixed by not relying on `audfactory.url()` to create url as it converts `.` to `/`.

`audfactory.url()` is still used in `Artfactory.ls()` and `Artifactory.glob()`, but these functions will be changed / removed anyway, see #46

### DIFF
--- a/audbackend/core/artifactory.py
+++ b/audbackend/core/artifactory.py
@@ -67,14 +67,9 @@ class Artifactory(Backend):
 
         """
         folder, file = self.split(path)
-        name, ext = utils.splitext(file, ext)
+        name, _ = utils.splitext(file, ext)
 
-        path = audfactory.url(
-            self.host,
-            repository=self.repository,
-            group_id=audfactory.path_to_group_id(folder),
-            name=name,
-        )
+        path = f'{self.host}/{self.repository}/{folder}/{name}'
 
         return path
 
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -305,6 +305,12 @@ def test_exists(tmpdir, backend):
             None,
         ),
         (
+            os.path.join('dir.to', 'file.ext'),
+            'dir.to/file.ext',
+            '1.0.0',
+            None,
+        ),
+        (
             os.path.join('dir', 'to', 'file.ext'),
             'alias.ext',
             '1.0.0',
@@ -341,12 +347,7 @@ def test_file(tmpdir, local_file, remote_file, version, ext, backend):
     assert not backend.exists(remote_file, version, ext=ext)
     backend.put_file(local_file, remote_file, version, ext=ext)
     # operation will be skipped
-    backend.put_file(
-        local_file,
-        remote_file,
-        version,
-        ext=ext,
-    )
+    backend.put_file(local_file, remote_file, version, ext=ext)
     assert backend.exists(remote_file, version, ext=ext)
 
     backend.get_file(remote_file, local_file, version, ext=ext)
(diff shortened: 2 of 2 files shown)


audbackend/core/artifactory.py
```diff
@@ -67,14 +67,9 @@ def _folder(
...
-        name, ext = utils.splitext(file, ext)
+        name, _ = utils.splitext(file, ext)
 
-        path = audfactory.url(
-            self.host,
-            repository=self.repository,
-            group_id=audfactory.path_to_group_id(folder),
-            name=name,
-        )
+        path = f'{self.host}/{self.repository}/{folder}/{name}'
```
hagenw, line 72: It has the nice side effect that it is no also easier to see what the result will be.

tests/test_backend.py
```diff
@@ -356,11 +357,6 @@ def test_file(tmpdir, local_file, remote_file, version, ext, backend):
     backend.remove_file(remote_file, version, ext=ext)
     assert not backend.exists(remote_file, version, ext=ext)
 
-    if ext is None:
-        _, ext = os.path.splitext(local_file)
-    else:
-        ext = '.' + ext
```
hagenw, line 362: Why can you remove this part?

frankenjoe: It's left over when we were testing if the path returned by `put_file()` has the expected extension. But since we no longer return a path, these lines are also no longer needed. See https://github.com/audeering/audbackend/pull/47/files#diff-31e088e0e9d455d2164fd6251e7c23a77afc5561b3746cdf83e08d6a05142154L349

Issue #48, opened 2022-12-07
Add check for path characters in Backend.ls()


hagenw:
When `Backend.ls()` is updated we should also add a check for its argument `path` that it contains only valid chars, the same as we do for all other methods.

frankenjoe:
Done in #53

PR #53, opened 2022-12-07
Reimplement Backend.ls()

### LINKED ISSUE #10: Replace _glob() with a simpler function

### LINKED ISSUE #37: ls() does not raise expected error

### LINKED ISSUE #49: Same virtual path can translate into different backend paths

frankenjoe:
Relates to #46 

Closes https://github.com/audeering/audbackend/issues/10 #25  #37 https://github.com/audeering/audbackend/issues/48 #49

image

This re-implements `Backend.ls()` by not returning the path on the backend, but a list of virtual paths together with version and extension, so we can close #10 and #25 

In the test we now create three files that have the same virtual path, but we set a different extension. If a backend does not translate it into three different paths, the test will fail, so we can close #49 

We now raise a `FileNotFoundError` error  folder does not exist, so we can close #37

### DIFF
--- a/audbackend/core/artifactory.py
+++ b/audbackend/core/artifactory.py
@@ -125,7 +125,7 @@ class Artifactory(Backend):
 
         # <host>/<repository>/<folder>/<name>/<version>/<name>-<version><ext>
         # ->
-        # (<folder>/<name><ext>, <version>, <ext>)
+        # (<folder>/<name><ext>, <ext>, <version>)
 
         result = []
         for full_path in paths:
@@ -142,7 +142,7 @@ class Artifactory(Backend):
             ext = file[len(name) + len(version) + 1:]
             path = self.join(folder, f'{name}{ext}')
 
-            result.append((path, version, ext))
+            result.append((path, ext, version))
 
         return result
 
--- a/audbackend/core/artifactory.py
+++ b/audbackend/core/artifactory.py
@@ -69,7 +69,10 @@ class Artifactory(Backend):
         folder, file = self.split(path)
         name, _ = utils.splitext(file, ext)
 
-        path = f'{self.host}/{self.repository}/{folder}/{name}'
+        if folder:
+            path = f'{self.host}/{self.repository}/{folder}/{name}'
+        else:
+            path = f'{self.host}/{self.repository}/{name}'
 
         return path
 
@@ -109,15 +112,40 @@ class Artifactory(Backend):
 
     def _ls(
             self,
-            path: str,
+            folder: str,
     ):
-        r"""List content of path."""
-        path = audfactory.url(
-            self.host,
-            repository=self.repository,
-            group_id=audfactory.path_to_group_id(path),
-        )
-        return [p.name for p in audfactory.path(path)]
+        r"""List content of folder."""
+        root = self._folder(folder, '')
+
+        path = audfactory.path(root)
+        try:
+            paths = [str(x) for x in path.glob("**/*") if x.is_file()]
+        except self._non_existing_path_error:  # pragma: nocover
+            paths = []
+
+        # <host>/<repository>/<folder>/<name>/<version>/<name>-<version>.<ext>
+        # ->
+        # (<folder>/<name>.<ext>, <version>, <ext>)
+
+        result = []
+        for full_path in paths:
+
+            host_repo = f'{self.host}/{self.repository}'
+            full_path = full_path[len(host_repo) + 1:]  # remove host and repo
+
+            tokens = full_path.split(os.path.sep)
+            file = tokens[-1]
+            version = tokens[-2]
+            name = tokens[-3]
+            folder = os.path.sep.join(tokens[:-3])
+            ext = file[len(name) + len(version) + 1:]
+            path = os.path.join(folder, f'{name}{ext}')
+            if ext:
+                ext = ext[1:]  # remove .
+
+            result.append((path, version, ext))
+
+        return result
 
     def _path(
             self,
--- a/audbackend/core/artifactory.py
+++ b/audbackend/core/artifactory.py
@@ -116,7 +116,7 @@ class Artifactory(Backend):
     ):
         r"""List all files under folder.
 
-        Returns an empty list if folder does not exist.
+        Return an empty list if no files match or folder does not exist.
 
         """
         root = self._folder(folder, '')
--- a/audbackend/core/artifactory.py
+++ b/audbackend/core/artifactory.py
@@ -132,14 +132,15 @@ class Artifactory(Backend):
 
             host_repo = f'{self.host}/{self.repository}'
             full_path = full_path[len(host_repo) + 1:]  # remove host and repo
+            full_path = full_path.replace('/', self.sep)
+            tokens = full_path.split('/')
 
-            tokens = full_path.split(os.path.sep)
             file = tokens[-1]
             version = tokens[-2]
             name = tokens[-3]
-            folder = os.path.sep.join(tokens[:-3])
+            folder = self.sep.join(tokens[:-3])
             ext = file[len(name) + len(version) + 1:]
-            path = os.path.join(folder, f'{name}{ext}')
+            path = self.join(folder, f'{name}{ext}')
             if ext:
                 ext = ext[1:]  # remove .
 
--- a/audbackend/core/artifactory.py
+++ b/audbackend/core/artifactory.py
@@ -114,7 +114,11 @@ class Artifactory(Backend):
             self,
             folder: str,
     ):
-        r"""List all files under folder."""
+        r"""List all files under folder.
+
+        Returns an empty list if folder does not exist.
+
+        """
         root = self._folder(folder, '')
 
         path = audfactory.path(root)
--- a/audbackend/core/artifactory.py
+++ b/audbackend/core/artifactory.py
@@ -61,7 +61,7 @@ class Artifactory(Backend):
     ) -> str:
         r"""Convert to backend folder.
 
-        <folder>/<name>.<ext>
+        <folder>/<name><ext>
         ->
         <host>/<repository>/<folder>/<name>/
 
@@ -114,7 +114,7 @@ class Artifactory(Backend):
             self,
             folder: str,
     ):
-        r"""List content of folder."""
+        r"""List all files under folder."""
         root = self._folder(folder, '')
 
         path = audfactory.path(root)
@@ -123,9 +123,9 @@ class Artifactory(Backend):
         except self._non_existing_path_error:  # pragma: nocover
             paths = []
 
-        # <host>/<repository>/<folder>/<name>/<version>/<name>-<version>.<ext>
+        # <host>/<repository>/<folder>/<name>/<version>/<name>-<version><ext>
         # ->
-        # (<folder>/<name>.<ext>, <version>, <ext>)
+        # (<folder>/<name><ext>, <version>, <ext>)
 
         result = []
         for full_path in paths:
@@ -141,8 +141,6 @@ class Artifactory(Backend):
             folder = self.sep.join(tokens[:-3])
             ext = file[len(name) + len(version) + 1:]
             path = self.join(folder, f'{name}{ext}')
-            if ext:
-                ext = ext[1:]  # remove .
 
             result.append((path, version, ext))
 
--- a/audbackend/core/backend.py
+++ b/audbackend/core/backend.py
@@ -332,18 +332,18 @@ class Backend:
     ) -> typing.List[typing.Tuple[str, str, str]]:
         r"""List all files under folder.
 
-        For each match,
-        path, version and extension is returned.
-        If extension is not empty,
+        Returns a sorted list of tuples
+        with path, version and file extension.
+        If file extension is not empty,
         it starts with a dot.
         If folder does not exist,
-        the result is an empty list.
+        an empty list is returned.
 
         Args:
             folder: folder on backend
 
         Returns:
-            sorted list of tuples (path, version, extension)
+            list of tuples
 
         Raises:
             ValueError: if ``folder`` contains invalid character
--- a/audbackend/core/backend.py
+++ b/audbackend/core/backend.py
@@ -335,7 +335,7 @@ class Backend:
         r"""List all files under folder.
 
         Returns a sorted list of tuples
-        with path, version and file extension.
+        with path, extension and version.
         If file extension is not empty,
         it starts with a dot.
 
@@ -345,7 +345,7 @@ class Backend:
                 only include the latest
 
         Returns:
-            list of tuples
+            list of tuples (path, extension, version)
 
         Raises:
             FileNotFoundError: if ``folder`` does not exist
@@ -353,9 +353,9 @@ class Backend:
 
         Example:
             >>> backend.ls('folder')[:2]
-            [('folder/name.ext', '1.0.0', '.ext'), ('folder/name.ext', '2.0.0', '.ext')]
+            [('folder/name.ext', '.ext', '1.0.0'), ('folder/name.ext', '.ext', '2.0.0')]
             >>> backend.ls('folder', latest_version=True)[:1]
-            [('folder/name.ext', '2.0.0', '.ext')]
+            [('folder/name.ext', '.ext', '2.0.0')]
 
         """  # noqa: E501
         utils.check_path_for_allowed_chars(folder)
@@ -372,7 +372,7 @@ class Backend:
         if latest_version:
             # d[path, ext] = ['1.0.0', '2.0.0']
             d = {}
-            for p, v, e in paths:
+            for p, e, v in paths:
                 key = (p, e)
                 if key not in d:
                     d[key] = []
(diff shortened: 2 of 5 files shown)


hagenw:
Would it make sense to provide an option to list only the latest versions of all entries? Or is the use case for `ls()` typically to really access everything?

audbackend/core/artifactory.py
```diff
@@ -109,15 +112,39 @@ def _glob(
 
     def _ls(
             self,
-            path: str,
+            folder: str,
```
frankenjoe, line 115: We could set `''` as default value, then it would simply show all files on the backend if no `folder` is specified.

hagenw:
What is the motivation to return an empty list when the virtual path does exists instead of raising a `FileNotFoundError`,
which we also get with:

```python
>>> import os
>>> os.listdir('non-existent')
...
FileNotFoundError: [Errno 2] No such file or directory: 'non-existent'
```

frankenjoe:
> What is the motivation to return an empty list when the virtual path does exists instead

```python
audeer.list_file_names('does-not-exist')
```
```
[]
```

But on the other hand, we also have:

```python
audeer.list_dir_names('does-not-exist')
```
```
FileNotFoundError: [Errno 2] No such file or directory: 'does-not-exist'
```

frankenjoe:
> > Would it make sense to provide an option to list only the latest versions of all entries? Or is the use case for ls() typically to really access everything?
> 
> I also though this might be convenient. So yes let's add it then.

done

hagenw:
Inside `audeer` we don't state that `list_dir_names()` can raise an exception. We should add this or change its behavior. I'm still not decided what would be the best.

audbackend/core/artifactory.py
hagenw: OK, then I would set `'/'` as default.

frankenjoe:
> Inside `audeer` we don't state that `list_dir_names()` can raise an exception. We should add this or change its behavior. I'm still not decided what would be the best.

Yes, we should have the same behavior everywhere. For me returning an empty list seems fine, a folder that does not exist also has no content :)

audbackend/core/artifactory.py
frankenjoe: done

hagenw:
> > Inside `audeer` we don't state that `list_dir_names()` can raise an exception. We should add this or change its behavior. I'm still not decided what would be the best.
> 
> Yes, we should have the same behavior everywhere. For me returning an empty list seems fine, a folder that does not exist has no content :)

This is true, but it behaves differently then other programs do, e.g.
```bash
$ ls non-existent
ls: cannot access 'non-existent': No such file or directory
```

hagenw:
What is a bit unfortunate now is the following behavior:

```python
>>> backend.ls('folder/name.ext')
...
FileNotFoundError: [Errno 2] No such file or directory: 'folder/name.ext'
```

which is wrong as the file exists. Maybe we should return the single file in this case?

```python
>>> backend.ls('folder/name.ext')
[('folder/name.ext', '1.0.0', '.ext'), ('folder/name.ext', '2.0.0', '.ext')]
```

frankenjoe:
But then we also have to add an `ext` argument since unfortunately the way we implemented the `Artifactory` backend, different extensions can lead to a different paths on the backend, see #49.

frankenjoe:
> What is a bit unfortunate now is the following behavior:
> 
> ```python
> >>> backend.ls('folder/name.ext')
> ...
> FileNotFoundError: [Errno 2] No such file or directory: 'folder/name.ext'
> ```
> 
> which is wrong as the file exists

We now always append `/` to `folder`, so at least the error message would read:

```
FileNotFoundError: [Errno 2] No such file or directory: 'folder/name.ext/'
```

I think this way it's more obvious to the user why it fails.

hagenw:
If we remove `ext` I would again vote to support `backend.ls('folder/name.ext')` as `ls` under Linux does as well.

frankenjoe:
> Would you like to first remove `ext` and then continue here, or would you like to merge this first?

Let's finish this one first.

frankenjoe:
> If we remove `ext` I would again vote to support `backend.ls('folder/name.ext')` as `ls` under Linux does as well.

Agree, could you please create an issue that we don't forget it.

hagenw:
I created https://github.com/audeering/audbackend/issues/59

audbackend/core/backend.py
```diff
@@ -321,28 +321,71 @@ def latest_version(
...
+            folder: str = '/',
+            *,
+            latest_version: bool = False,
+    ) -> typing.List[typing.Tuple[str, str, str]]:
+        r"""List all files under folder.
+
+        Returns a sorted list of tuples
+        with path, extension and version.
+        If file extension is not empty,
+        it starts with a dot.
```
hagenw, line 340: It's not obvious what `it` is referring to. Looking only at the sentence one would guess `sorted list of tuples`, but this makes no sense.

audbackend/core/backend.py
```diff
@@ -321,28 +321,71 @@ def latest_version(
...
+        """  # noqa: E501
+        utils.check_path_for_allowed_chars(folder)
+        if not folder.endswith('/'):
+            folder += '/'
+        paths = self._ls(folder)
+        paths = sorted(paths)
+
+        if len(paths) == 0:
+            if folder == '/':  # pragma: no cover
+                # special case that there are no files on the backend
```
hagenw, line 369: Sounds like we should be able to create a test for it, e.g. creating a new empty repo?

audbackend/core/backend.py
```diff
@@ -321,28 +321,71 @@ def latest_version(
...
+                return []
+            else:
+                utils.raise_file_not_found_error(folder)
+
+        if latest_version:
+            # d[path, ext] = ['1.0.0', '2.0.0']
```
hagenw, line 375:
```suggestion
            # d[(path, ext)] = ['1.0.0', '2.0.0']
```

audbackend/core/backend.py
```diff
@@ -321,28 +321,71 @@ def latest_version(
...
+            d = {}
+            for p, e, v in paths:
+                key = (p, e)
+                if key not in d:
+                    d[key] = []
+                d[key].append(v)
+            # d[path, ext] = '2.0.0'
```
hagenw, line 382:
```suggestion
            # d[(path, ext)] = '2.0.0'
```

audbackend/core/backend.py
```diff
@@ -321,28 +321,71 @@ def latest_version(
...
+            for key, vs in d.items():
+                d[key] = audeer.sort_versions(vs)[-1]
+            # [(path, ext, '2.0.0'), ...]
```
hagenw, line 385: Maybe change to

```suggestion
            # [(path, ext, '2.0.0')]
```
as for me it was not obvious if `...` refer to other versions of the same file or to other (path, ext) combinations.

tests/test_backend.py
```diff
@@ -462,26 +454,62 @@ def test_glob(tmpdir, files, pattern, folder, expected, backend):
...
-    assert backend.ls(backend_path) == expected_content
+    # test
+
+    for folder, expected, expected_latest in [
+        ('/', content, content_latest),
+        ('sub', sub_content, sub_content_latest),
+    ]:
+        folder = backend.join(prefix, folder)
+
+        expected = [  # replace ext where it is None
```
hagenw, line 500: Maybe

```suggestion
        expected = [  # remove ext where it is None
```

tests/test_backend.py
```diff
@@ -462,26 +454,62 @@ def test_glob(tmpdir, files, pattern, folder, expected, backend):
...
+            (p, f".{p.split('.')[-1]}" if e is None else e, v)
+            for p, e, v in expected
+        ]
+        expected = sorted(expected)
+        assert backend.ls(folder) == expected
+
+        expected_latest = [  # replace ext where it is None
```
hagenw, line 507: Maybe

```suggestion
        expected_latest = [  # remove ext where it is None
```

audbackend/core/artifactory.py
```diff
@@ -109,15 +112,39 @@ def _glob(
...
-        )
-        return [p.name for p in audfactory.path(path)]
+        r"""List all files under folder."""
+        root = self._folder(folder, '')
+
+        path = audfactory.path(root)
+        try:
+            paths = [str(x) for x in path.glob("**/*") if x.is_file()]
+        except self._non_existing_path_error:  # pragma: nocover
+            paths = []
```
hagenw, line 124: What case is this?

If the path does not exist it should raise an error.
If this covers the case of an empty repo with the `/` request, we should add a test for it.

hagenw:
image

Looks like an all-time record ;)

audbackend/core/backend.py
frankenjoe: It refers to `file extension`, but since we anyway plan to get rid of it in #58 I simply removed the sentence.

tests/test_backend.py
frankenjoe: If the file extension is `None` we replace it with the default file extension (i.e. characters after last dot), so I think the term "replace" is correct here.

tests/test_backend.py
frankenjoe: See previous comment.

audbackend/core/artifactory.py
frankenjoe: The empty repo case is covered in `Backend.ls()`. So yes, it's the case that the folder does not exist. We then return an empty list and an error is raised in `Backend.ls()`, this way we avoid that the implementation is responsible for raising the expected error. I now added a comment to the docstring that an empty list should be returned by `_ls()` if the folder does not exist.

frankenjoe:
> Looks like an all-time record ;)

Hehehe, it was actually the function to list the files on a backend that kicked-off all those issues we are currently dealing with.

audbackend/core/backend.py
hagenw: Instead of rewirting the tests, you could add an additional test that creates a new empty repo just for this particular test?

tests/test_backend.py
hagenw: You are right.
I just tought `p.split('.')[-1]` looks not like replacement, but if you read carefully you understand it.

audbackend/core/artifactory.py
hagenw: How do you differentiate an empty list returned by a non-existent folder from an empty list returned by an empty folder?

frankenjoe: Since there is no function to create a folder, empty folders cannot exist on the backend. In reality they can of course, but I think it still makes sense to raise always raise an error. Except for the special case of `/`, for which we return an empty list if no file exists yet in the repo.

frankenjoe: So in other words: a folder only exists if it contains at least one file.

hagenw: I'm not that excited about this solution as a user might find other ways to delete all files in a folder, but the folder still exists. I suppose that `backend.exist(folder)` will also return `True` then, and `backend.ls(folder)` will raise an error that it does not exist.

If we want to stay with the behavior that the backend should return a special value we might think of changing the value, e.g. to `None`.

frankenjoe: Another reason for always raising an error if no files are found is that this will guarantee the same behavior on every backend. E.g. assume we put the same file on backend A and B and remove it afterwards. Now assume on backend A removes the folder (or maybe a concept like folders does not even exist), but on backend B we the folder is kept. If we return an empty list if the folder exists, A will now return an empty list, but B raise an error.

audbackend/core/backend.py
frankenjoe: I anyway wanted to rewrite the tests to ensure the backend is cleaned after each test. This is now automatically done by the `backend` fixture. I now added a test for calling `ls()` on an empty repository and removed the pragma comment.

frankenjoe: I also simplified the tests in a couple of places by using `audeer.touch()` and not encoding the unique ID in the path but already in the repository.

frankenjoe: And we now keep the list of backends that will be tested in `conftest.py`. So during local debugging the developer can comment out certain backends to speed up the tests.

audbackend/core/artifactory.py
hagenw: Yes, makes sense. I forgot that backends don't need support folders, and running `backend.exists()` on a folder would indeed not make any sense.

The only problematic part is that it then still might be useful to stay with our `audfactory` package as this provides an option to the user to check if folders exist on Artifactory and remove them.

hagenw: Is the behavior in `Backend.versions()` that you get `[]` if the folder is empty and a `FileNotFoundError` if the folder does not exist?

frankenjoe: No, currently we never raise an error but return an empty list if the file does not exist.

tests/test_backend.py
```diff
@@ -391,108 +329,95 @@ def test_file(tmpdir, local_file, remote_file, version, ext, backend):
...
-def test_ls(tmpdir, path, content, expected_content, backend):
+def test_ls(tmpdir, backend):
 
-    backend_path = backend.join(
-        pytest.ID,
-        'test_ls',
-        path,
-    )
+    assert backend.ls() == []
+    assert backend.ls('/') == []
```
hagenw, line 370: This feels still a little bit strange, but I guess `/`, the root of the backend,  is officially not a folder and it makes sense to return an empty list and not raise a `FileNotFoundError`. I think we should add a section to the docstring, stating that this is the only case you can get an empty list?

audbackend/core/backend.py
```diff
@@ -321,28 +321,73 @@ def latest_version(
...
-    ) -> typing.List:  # pragma: no cover
-        r"""List content of path."""
+            folder: str,
+    ) -> typing.List[typing.Tuple[str, str, str]]:  # pragma: no cover
+        r"""List all files under folder.
+
+        Returns an empty list if folder does not exist.
+
+        """
         raise NotImplementedError()
```
hagenw, line 331: We need to add documentation that a developer of a new backend should return an empty list for empty folders (if actual folders exist on the backend). Should we add a comment here or should we add a special developer section to the documentation?

tests/test_backend.py
frankenjoe: We now say:

image

audbackend/core/backend.py
frankenjoe: I changed the comment to: `Return an empty list if no files match or folder does not exist.` 

And yes, I plan to include a usage section where we discuss how to implement a custom backend.

hagenw: Good.

I also created https://github.com/audeering/audbackend/issues/60 to track the documentation part.

audbackend/core/artifactory.py
hagenw: I created https://github.com/audeering/audbackend/issues/61 to track the proposed change for `backend.versions()`

PR #54, opened 2022-12-08
DOC: add docstring examples

frankenjoe:
image

image

image

image

image

image

image

image

image

image

image

image

image

image

### DIFF
--- a/audbackend/core/api.py
+++ b/audbackend/core/api.py
@@ -33,6 +33,11 @@ def create(
     Raises:
         ValueError: if registry name does not exist
 
+    Example:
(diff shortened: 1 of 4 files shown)


frankenjoe:
Docstring example for `Backend.ls` and `create()` I will add once #53 is done.

hagenw:
Great, thanks a lot.

I think you could also add an example to `audbackend.register()`:

```python
>>> register('file-system', audbackend.FileSystem)
```

Maybe we should also provide a function to list registered backends?

hagenw:
I created https://github.com/audeering/audbackend/issues/55 to track this.

PR #56, opened 2022-12-08
Add available() to list available backends

### LINKED ISSUE #55: Add audbackend.available()
We have `audbackend.register()` to register backend classes to backend names. But at the moment there is now way to check which names (and corresponding classes) exist already. We could add `audbackend.available()` to provide this information.

frankenjoe:
Closes #55 

image

Note that `Backend.__repr__()` is now implemented to have a nicer representation of the backend objects. The docstring example of `Backend.create()` has been improved accordingly.

### DIFF
--- a/audbackend/core/api.py
+++ b/audbackend/core/api.py
@@ -18,8 +18,14 @@ r"""Backend registry."""
 def available() -> typing.Dict[str, typing.List[Backend]]:
     r"""List available backends.
 
+    Returns a dictionary with
+    registered backend names as keys
+    (see :func:`audbackend.register`)
+    and a list with backend instances as values
+    (see :func:`audbackend.create`).
+
     Returns:
-        sorted dictionary with backends
+        dictionary with backends
 
     Examples:
         >>> list(available())
--- a/audbackend/core/api.py
+++ b/audbackend/core/api.py
@@ -31,9 +31,9 @@ def available() -> typing.Dict[str, typing.List[Backend]]:
         >>> list(available())
         ['artifactory', 'file-system']
         >>> available()['artifactory']
-        [('Artifactory', 'https://audeering.jfrog.io/artifactory', 'repo')]
+        [('audbackend.core.artifactory.Artifactory', 'https://host.com', 'repo')]
 
-    """
+    """  # noqa: E501
     result = {}
 
     for name in sorted(backend_registry):
@@ -67,10 +67,10 @@ def create(
     Example:
         >>> create(
         ...     'artifactory',
-        ...     'https://audeering.jfrog.io/artifactory',
+        ...     'https://host.com',
         ...     'repo',
         ... )
-        ('Artifactory', 'https://audeering.jfrog.io/artifactory', 'repo')
+        ('audbackend.core.artifactory.Artifactory', 'https://host.com', 'repo')
 
     """
     if name not in backend_registry:
--- a/audbackend/core/api.py
+++ b/audbackend/core/api.py
@@ -15,7 +15,7 @@ backend_registry = {
 r"""Backend registry."""
 
 
-def available() -> typing.Dict[str, typing.List[typing.Tuple[str, str]]]:
+def available() -> typing.Dict[str, typing.List[Backend]]:
     r"""List available backends.
 
     Returns:
--- a/audbackend/core/api.py
+++ b/audbackend/core/api.py
@@ -22,12 +22,6 @@ def available() -> typing.Dict[str, typing.List[Backend]]:
         sorted dictionary with backends
 
     Examples:
-        >>> create(
-        ...     'artifactory',
-        ...     'https://audeering.jfrog.io/artifactory',
-        ...     'repo',
-        ... )
-        ('Artifactory', 'https://audeering.jfrog.io/artifactory', 'repo')
         >>> list(available())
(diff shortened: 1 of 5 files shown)


hagenw:
Interesting, this is not exactly what I had in mind in #55.
Seems we have two different layers here.

First we have registered backends (which I had in mind):
```python
>>> audbackend.available()
[('artifactory', 'audbackend.Artifactory'), ('file-system', 'audbackend.FileSystem')]
```
and then we have the actual created backends (what is implemented in this pull request)
```python
>>> audbackend.available()
{'artifactory': ...}
```

Both are different and both might be interesting to know for a user. So maybe we need two different functions, or should we provide an argument to `available()`?

frankenjoe:
Currently, even if no backends are created it still lists the registered names:

```python
audbackend.available()
```
```
{'artifactory': [], 'file-system': []}
```

With this information the user can now do:

```python
backend = audbackend.create('artifactory', ...)
```

So I don't think we need to provide another function.

audbackend/core/api.py
```diff
@@ -15,6 +15,31 @@
...
+    r"""List available backends.
+
+    Returns:
+        sorted dictionary with backends
+
+    Examples:
+        >>> list(available())
+        ['artifactory', 'file-system']
+        >>> available()['artifactory']
+        [('Artifactory', 'https://audeering.jfrog.io/artifactory', 'repo')]
```
frankenjoe, line 28: Maybe we should change the output:

```
[('audbackend.Artifactory', 'https://audeering.jfrog.io/artifactory', 'repo')]
```

?

hagenw:
```python
>>> available()['artifactory']
[('Artifactory', 'https://audeering.jfrog.io/artifactory', 'repo')]
```
I think we can only register one backend per string (at least `audbackend.register()` states that it will overwrite existing ones), so I guess the following would be sufficient:
```python
>>> available()['artifactory']
('Artifactory', 'https://audeering.jfrog.io/artifactory', 'repo')
```

hagenw:
> Currently, even if no backends are created it still lists the registered names:
> 
> ```python
> audbackend.available()
> ```
> 
> ```
> {'artifactory': [], 'file-system': []}
> ```
> 
> With this information the user can now do:
> 
> ```python
> backend = audbackend.create('artifactory', ...)
> ```
> 
> So I don't think we need to provide another function.

Good point, wasn't ware of this.
Would be great to discuss this in the docstring, e.g.

```
It returns a dictionary with all registered backend names as keys,
compare :func:`audbackend.register`.
If a backend instance exist for a registered backend,
it is listed as the value of the dictionary,
compare :func:`audbackend.create`.
```

frankenjoe:
> I think we can only register one backend per string (at least audbackend.register() states that it will overwrite existing ones), so I guess the following would be sufficient:

For the same registered backend class we can still have several objects with different host / repos:

```python
audbackend.create(
    'file-system',
    'host',
    'repo',
)
audbackend.create(
    'file-system',
    'another-host',
    'another-repo',
)
audbackend.available()['file-system']
```
```
[('FileSystem', '/media/jwagner/Data/Git/pyaudbackend/my/host', 'repo'), ('FileSystem', '/media/jwagner/Data/Git/pyaudbackend/my/another-host', 'another-repo')]
```

 Otherwise we would not be able to load e.g. `emodb` from our public server but other databases from our internal one.

frankenjoe:
> Good point, wasn't ware of this.
> Would be great to discuss this in the docstring, e.g.

I added:

image

audbackend/core/api.py
frankenjoe: You are right, we either stay with only the class name or we expand to the full module name, i.e.:

```
[('audbackend.core.artifactory.Artifactory', 'https://audeering.jfrog.io/artifactory', 'repo')]
```

frankenjoe: Since we may have two implementations with the same class name in two different modules, I decided to change to the full name. We now have:

image

and 

image

audbackend/core/api.py
```diff
@@ -15,6 +15,37 @@
...
+    Returns a dictionary with
+    registered backend names as keys
+    (see :func:`audbackend.register`)
+    and a list with backend instances as values
+    (see :func:`audbackend.create`).
+
+    Returns:
+        dictionary with backends
+
+    Examples:
```
hagenw, line 30: At the moment we always use

```suggestion
    Example:
```

In the future I would like to change it to `Examples` for all entries, but for now I would suggest to stick to what we do in other docstrings.

hagenw:
> > I think we can only register one backend per string (at least audbackend.register() states that it will overwrite existing ones), so I guess the following would be sufficient:
> 
> For the same registered backend class we can still have several objects with different host / repos:
> 
> ```python
> audbackend.create(
>     'file-system',
>     'host',
>     'repo',
> )
> audbackend.create(
>     'file-system',
>     'another-host',
>     'another-repo',
> )
> audbackend.available()['file-system']
> ```
> 
> ```
> [('FileSystem', 'host', 'repo'), ('FileSystem', 'another-host', 'another-repo')]
> ```
> 
> Otherwise we would not be able to load e.g. `emodb` from our public server but other databases from our internal one.

Ah, yes, so I agree that it makes sense to stay with a list as return type.

PR #66, opened 2023-01-03
Add support for Python 3.10

hagenw:
Add a test for Python 3.10 under Ubuntu and mention Python 3.10 as supported version in `setup.cfg`


hagenw:
After https://github.com/audeering/audfactory/pull/45 is merged and we have a new release of `audfactory`, we could require that and don't have to specify the `dohq-artifactory` version here.

PR #82, opened 2023-04-13
Force archive extension and add support for TAR.GZ

### LINKED ISSUE #78: Consider adding file extension to `Backend.put_archive()`
At the moment our call to `audbackend.Backend.put_archive()` is:

```python
backend.put_archive(tmp, files, 'folder/name', '2.0.0')
```

which is slightly different from all the other functions that require a user to provide the file extension as well with the name.
In this case it would be:

```python
backend.put_archive(tmp, files, 'folder/name.zip', '2.0.0')
```

One of the reasons we didn't included it here was that the function can only create a ZIP archive at the moment,
which means if we add support for `folder/name.zip` we have to raise an error if another file extension is provided instead.

frankenjoe:
Closes #78 

The methods `Backend.put_archive()` and `Backend.get_archive()` do no longer extend the archive name with `.zip`. Instead the user has to provide the extension. In addition to ZIP, also TAR.GZ is now supported.

image

image

### DIFF
--- a/audbackend/core/backend.py
+++ b/audbackend/core/backend.py
@@ -403,8 +403,7 @@ class Backend:
                 Only folders and files below ``src_root``
                 will be included into the archive
             files: relative path to file(s) from ``src_root``
-            dst_path: path to archive on backend,
-                e.g. ``sub/archive.zip``
+            dst_path: path to archive on backend
             version: version string
             tmp_root: directory under which archive is temporarily created.
                 Defaults to temporary directory of system
--- a/audbackend/core/backend.py
+++ b/audbackend/core/backend.py
@@ -114,7 +114,7 @@ class Backend:
         r"""Get archive from backend and extract.
 
         The archive type is derived from the extension of ``src_path``.
-        See :func:`audeer.create_archive` for supported extensions.
+        See :func:`audeer.extract_archive` for supported extensions.
 
         Args:
             src_path: path to archive on backend
@@ -132,6 +132,7 @@ class Backend:
             FileNotFoundError: if ``tmp_root`` does not exist
             ValueError: if ``src_path`` contains invalid character
             RuntimeError: if extension of ``src_path`` is not supported
+            RuntimeError: if ``src_path`` is a malformed archive
 
         Examples:
             >>> dst_root = audeer.path(tmp, 'dst')
--- a/audbackend/core/backend.py
+++ b/audbackend/core/backend.py
@@ -111,11 +111,11 @@ class Backend:
             tmp_root: str = None,
             verbose: bool = False,
     ) -> typing.List[str]:
-        r"""Get archive from backend and extract.
+        r"""Get ZIP archive from backend and extract.
 
         Args:
-            src_path: path to archive on backend,
-                e.g. ``media/archive1.zip``.
+            src_path: path to ZIP archive on backend,
+                e.g. ``sub/archive.zip``.
                 If path does not end on ``.zip``
                 it is automatically extended to ``<src_path>.zip``
             dst_root: local destination directory
@@ -390,7 +390,7 @@ class Backend:
             tmp_root: str = None,
             verbose: bool = False,
     ):
-        r"""Create archive and put on backend.
+        r"""Create ZIP archive and put on backend.
 
         The operation is silently skipped,
         if an archive with the same checksum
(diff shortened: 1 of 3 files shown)


hagenw:
Maybe we should extent both docstring texts mentioning that it will create and load only ZIP archives?

frankenjoe:
> Yes, this would already be enough.

done

audbackend/core/backend.py
```diff
@@ -132,11 +134,12 @@ def get_archive(
...
         Examples:
             >>> dst_root = audeer.path(tmp, 'dst')
-            >>> backend.get_archive('folder/name', dst_root, '1.0.0')
+            >>> backend.get_archive('folder/name.zip', dst_root, '1.0.0')
             ['src.pth']
 
         """
-        src_path += '.zip'
+        if not src_path.endswith('.zip'):
+            src_path += '.zip'
```
hagenw, line 142: Would it make more sense to raise an error if the extension does not match?

I ask, because it can happen that you don't find an error if you make a small typo, e.g.

```python
>>> zip_file = 'archive.Zip'
>>> backend.put_archive('.', ['name.txt'], zip_file, '1.0.0')
>>> backend.exists(zip_file, '1.0.0')
False
```

frankenjoe: In that case we have to:

image

as mentioned in https://github.com/audeering/audbackend/issues/78#issuecomment-1506479093

hagenw: Yes, but we have to update all of them anyway as the `ext` argument is removed and we have breaking changes.

frankenjoe: In `audb` we do not use `ext` and also not `glob()` . But I just checked and in `audb.available()` we use `ls()`, which now has a new return type. So we indeed will break all packages anyway. I.e. it would indeed not matter if we add another breaking change on top.

hagenw: I'm not completely sure what is the best option:

* Automatically add `.zip` if file extension does not match. Disadvantage: `.Zip` typos might be unnoticed
* Raise an error if file extension is not `.zip`. Disadvantage: user might wonder why she has to provide a file extension, but cannot choose a different one
* Don't provide file extension and always append `.zip`. Disadvantage: feels unintuitive as we provide file extension to all other methods, e.g. `backend.exists('archive.zip', '1.0.0')`
* Allow for arbitrary or no file extension. Disadvantage: ZIP files can be stored on the backend with "wrong" file extensions and might not be supported by backend functionality like web UIs.

So it seems everything has a disadvantage ;)

frankenjoe: In `audeer` we currently support ZIP and TAR.GZ and raise an error if the archive name does not end on either `zip` or `tar.gz`:

image

So maybe we should indeed do the same here.

frankenjoe: Ok, let's simply not check for the file extension and let `audeer` do the work for us. In the docstring we then state that we support the archives as `audeer.create_archive()` and that an error is raised if a non-supported file extension is used. This also has the advantage that we will automatically support all archives types that can be handeled by `audeer`:

hagenw: OK, at least we then have another example with `audeer` where we do the same. So yes, select 2. and raise an error.

I guess more general programs/modules simply allow for arbitrary file extensions when creating the ZIP file, but as we specifically target backend it might be a good idea to force the correct name as you also indicated.

hagenw: O, let's go with your suggestion and let `audeer` handle it.

frankenjoe: done

hagenw: Cool, thanks.

tests/test_backend.py
```diff
@@ -22,32 +22,48 @@ def backend(request):
...
+            'tmp',
+        ),
+        # extension not supported
+        pytest.param(
+            ['file.ext', 'dir/to/file.ext'],
+            'not-supported.bad',
+            'group',
+            '2.0.0',
+            'tmp',
+            marks=pytest.mark.xfail(raises=RuntimeError),
```
hagenw, line 65: I think this tests only the wrong extension when trying to put an archive to the backend, not getting it. For the later you need to add something like:

```python
audeer.touch('archive.yip')
backend.put_file('archive.yip', 'archive.yip', '1.0.0')
backend.get_archive('archive.yip', 'download', version='1.0.0')
```

PR #83, opened 2023-04-13
Add pattern argument to Backend.ls()

frankenjoe:
Adds pattern argument to `Backend.ls()` to filter results with [fnmatch](https://docs.python.org/3/library/fnmatch.html#fnmatch.fnmatch).

image

In order to fit the docstring example I had to rearrange the test files, which also affects some of the other docstring examples.

### DIFF
--- a/audbackend/core/backend.py
+++ b/audbackend/core/backend.py
@@ -1,3 +1,4 @@
+import fnmatch
 import os
 import tempfile
 import typing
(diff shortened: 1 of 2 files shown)


hagenw:
Maybe add one small test for it, so we do not only rely on the docstring example for code coverage of the new line.

frankenjoe:
> Maybe add one small test for it, so we do not only rely on the docstring example for code coverage of the new line.

done

Issue #84, opened 2023-04-13
Should Backend.latest_version() raise FileNotFoundError?


hagenw:
Most of the backend methods raise `FileNotFoundError` when a `path` argument is provided that does not exist on the server.
`Backend.latest_version()` raises `RuntimeError` instead. One difference is that `latest_version()` does not get `version` and does hence not look for a single file.

frankenjoe:
Ah good you raise that issue. I think `FileNotFoundError` makes more sense. And I have a related question: should maybe `Backend.versions()` should raise `FileNotFoundError` in the first place? At the moment it returns an empty list I think.

hagenw:
It might be easier to work with `[]` instead of having to catch different errors, because `Backend.versions()` might not only raise `FileNotFoundError` but also some permissions errors which can be anything across different backends at the moment. If we make sure that a backend would always raise the same error for no permissions, we could think about changing it, but until then we cannot expect that the users can know which error she has to catch.

frankenjoe:
Ok, but somehow we have the problem with `get_file()` and here we raise a `FileNotFoundError` even if the file exists but the user has not the right permission. So maybe we should keep it simple and always raise `FileNotFoundError` and not care about permission or other problems?

hagenw:
No, at the moment we always return `[]`, which I think is fine.
If we want to raise errors instead we need to make sure that all backends will raise the same error types for the same behavior, which I think is not possible. E.g. if a file does not exists on a backend for which you cannot login, you will not get a `FileNotFoundError`, whereas for a backend that does not require login you will get a `FileNotFoundError`.

hagenw:
> We can catch any error and then raise FileNotFoundError

Yes, that what I stated above. If we include those changes then we can update the behavior of `versions()`, but not before.

frankenjoe:
Regarding 2. - I just came across https://github.com/audeering/audbackend/issues/61 and there you propose to raise an error instead of returning an empty list.

frankenjoe:
As all other functions `Backend.latest_version()` will now raise `BackendError`, see #87

PR #85, opened 2023-04-13
TST: test for all Backend errors

hagenw:
This documents `PermissionError` for `Backend.get_archive()` and `Backend.get_file()` if the user has no local write permissions.

In addition, it adds tests for all errors we currently have documented in `audbackend.Backend` methods,
and fixes a few typos in the docstrings.

### DIFF
--- a/audbackend/core/backend.py
+++ b/audbackend/core/backend.py
@@ -130,9 +130,11 @@ class Backend:
         Raises:
             FileNotFoundError: if archive does not exist on backend
             FileNotFoundError: if ``tmp_root`` does not exist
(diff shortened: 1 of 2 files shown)


frankenjoe:
> This documents `PermissionError` for `Backend.get_archive()` and `Backend.get_file()` if the user has no local write permissions.

In #84 we started a discussion if we want to raise `PermissionError` or always `FileNotFoundError` (or maybe a custom error saying that either the file doe snot exist or the user is lacking permission)  since we cannot know what errors are raised by our backends.

frankenjoe:
> In this pull request I'm only handling local permission erros.

Ok, makes sense.

Issue #86, opened 2023-04-14
Should ls('/') raise an error if repository does not exist?


frankenjoe:
We decided that `ls('/')` should return an empty list if there are no files in the repository yet. But currently we also return an empty list if the repository does not even exist.

```python
backend = audbackend.create(
    'file-system',
    'host',
    'does-not-exist',
)
backend.ls('/')
```
```
[]
```

I wonder if we should raise an error in that case?

hagenw:
The question is who should be responsible for creating the repo in the first place.
E.g. you might argue that `audbackend.create()` should raise the error already if the repo does not exist.
On the other hand for the file-system backend it might also be handy if the repo is generated on the fly when the first file is added, which seems to happen anyway at the moment:

```python
backend = audbackend.create(
    'file-system',
    'host',
    'does-not-exist',
)
path = audeer.touch('file.txt')
backend.put_file(path, 'file.txt', '1.0.0')
backend.ls()
```
returns
```
[('file.txt', '1.0.0')]
```

To be in line with that behavior, maybe even `backend.ls()` should then create the repo and return an empty list?

The behavior should be identical for all backends, I guess we need to raise the error already for `audbackend.create()` if the repo does not exists or the user has no access rights to it.

frankenjoe:
> The behavior should be identical for all backends, I guess we need to raise the error already for `audbackend.create()` if the repo does not exists or the user has no access rights to it.

Yes the behavior should be identical for all backends and I think the repository should exist and not be created on the fly since this might not be always possible. So  `audbackend.create()` would indeed be the best place to raise an error if the repository does not exist. I see two options here:

1. Add a method that checks if the repository exists and that  every backend has to implement
2. Create and delete a temporary file in the repository

frankenjoe:
> On the other hand for the file-system backend it might also be handy if the repo is generated on the fly

A backend could still create the repository on the fly in the constructor.

frankenjoe:
We now create the repository with `audbackend.create()` and if this fails raise an error, see #91

PR #87, opened 2023-04-17
Raise generic backend error

frankenjoe:
Relates to #84 

Currently we raise a `FileNotFoundError` in (almost) every occasion where an error on the backend occurs. But this can be misleading as there can be various reasons why an operation on the backend fails, e.g. a temporary connection issue. 

Here we introduce the more generic `BackendError` as suggested in:

image

Furthermore, we encourage the backend to raise the native exceptions, e.g. checks if a file exists are removed and `_ls()` no longer should return an empty list if the folder does not exist.

image

### Example

Example how to access the original error raised by the backend:

```python
try:
    backend.checksum('does-not-exist', '1.0.0')
except audbackend.BackendError as ex:
    print(ex.exception)
```
```
[Errno 2] No such file or directory: '/my/host/does-not-exist/1.0.0/does-not-exist'
```

### DIFF
--- a/audbackend/core/backend.py
+++ b/audbackend/core/backend.py
@@ -55,7 +55,7 @@ class Backend:
             MD5 checksum
 
         Raises:
-            FileNotFoundError: if file does not exist on backend
+            BackendError: if an error is raised on the backend
             ValueError: if ``path`` contains invalid character
 
         Examples:
@@ -64,10 +64,14 @@ class Backend:
 
         """
         utils.check_path_for_allowed_chars(path)
-        if not self._exists(path, version):
-            utils.raise_file_not_found_error(path, version=version)
+        # if not self._exists(path, version):
+        #     utils.raise_file_not_found_error(path, version=version)
 
-        return self._checksum(path, version)
+        return utils.call_function_on_backend(
+            self._checksum,
+            path,
+            version,
+        )
 
     def _exists(
             self,
@@ -92,6 +96,7 @@ class Backend:
             ``True`` if file exists
 
         Raises:
+            BackendError: if an error is raised on the backend
             ValueError: if ``path`` contains invalid character
 
         Examples:
@@ -101,7 +106,11 @@ class Backend:
         """
         utils.check_path_for_allowed_chars(path)
 
-        return self._exists(path, version)
+        return utils.call_function_on_backend(
+            self._exists,
+            path,
+            version,
+        )
 
     def get_archive(
             self,
@@ -129,7 +138,7 @@ class Backend:
             extracted files
 
         Raises:
-            FileNotFoundError: if archive does not exist on backend
+            BackendError: if an error is raised on the backend
             FileNotFoundError: if ``tmp_root`` does not exist
             PermissionError: if the user lacks write permissions
                 for ``dst_path``
@@ -195,7 +204,7 @@ class Backend:
             full path to local file
 
         Raises:
-            FileNotFoundError: if ``src_path`` does not exist on backend
+            BackendError: if an error is raised on the backend
             PermissionError: if the user lacks write permissions
                 for ``dst_path``
             ValueError: if ``src_path`` contains invalid character
@@ -210,13 +219,25 @@ class Backend:
 
         """
         utils.check_path_for_allowed_chars(src_path)
-        if not self._exists(src_path, version):
-            utils.raise_file_not_found_error(src_path, version=version)
 
         dst_path = audeer.safe_path(dst_path)
-        audeer.mkdir(os.path.dirname(dst_path))
+        dst_root = os.path.dirname(dst_path)
 
-        self._get_file(src_path, dst_path, version, verbose)
+        audeer.mkdir(dst_root)
+        if (
+                not os.access(dst_root, os.W_OK) or
+                (os.path.exists(dst_path) and not os.access(dst_root, os.W_OK))
+        ):
+            msg = f"Permission denied: '{dst_path}'"
+            raise PermissionError(msg)
+
+        utils.call_function_on_backend(
+            self._get_file,
+            src_path,
+            dst_path,
+            version,
+            verbose,
+        )
 
         return dst_path
 
@@ -288,11 +309,7 @@ class Backend:
             self,
             folder: str,
     ) -> typing.List[typing.Tuple[str, str, str]]:  # pragma: no cover
-        r"""List all files under folder.
-
-        Return an empty list if no files match or folder does not exist.
-
-        """
+        r"""List all files under folder."""
         raise NotImplementedError()
 
     def ls(
@@ -323,6 +340,7 @@ class Backend:
             list of tuples (path, version)
 
         Raises:
+            BackendError: if an error is raised on the backend
             FileNotFoundError: if ``folder`` does not exist
             ValueError: if ``folder`` contains invalid character
 
@@ -340,15 +358,8 @@ class Backend:
         utils.check_path_for_allowed_chars(folder)
         if not folder.endswith('/'):
             folder += '/'
-        paths = self._ls(folder)
+        paths = utils.call_function_on_backend(self._ls, folder)
         paths = sorted(paths)
-
-        if len(paths) == 0:
-            if folder == '/':
-                # special case that there are no files on the backend
-                return []
-            else:
-                utils.raise_file_not_found_error(folder)
 
         if pattern:
             paths = [(p, v) for p, v in paths if fnmatch.fnmatch(p, pattern)]
@@ -399,6 +410,7 @@ class Backend:
             verbose: show debug messages
 
         Raises:
+            BackendError: if an error is raised on the backend
             FileNotFoundError: if one or more files do not exist
             FileNotFoundError: if ``tmp_root`` does not exist
             ValueError: if ``dst_path`` contains invalid character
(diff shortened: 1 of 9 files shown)


tests/test_backend.py
```diff
@@ -425,8 +424,10 @@ def test_file(tmpdir, src_path, dst_path, version, backend):
 )
 def test_ls(tmpdir, backend):
 
-    assert backend.ls() == []
```
frankenjoe, line 428: This is related to #86. In fact, so far the repositories we use in our tests are only created in the moment we put the first file on the backend. The reason this didn't raise an error before was that the backend was returning an empty list when something went wrong, which was actually hiding the original problem of a non-existing repository.

audbackend/core/backend.py
```diff
@@ -210,13 +217,25 @@ def get_file(
...
-        audeer.mkdir(os.path.dirname(dst_path))
-
-        self._get_file(src_path, dst_path, version, verbose)
+        dst_root = os.path.dirname(dst_path)
+
+        audeer.mkdir(dst_root)
+        if (
+                not os.access(dst_root, os.W_OK) or
+                (os.path.exists(dst_path) and not os.access(dst_root, os.W_OK))
+        ):  # pragma: no cover
```
frankenjoe, line 228: @hagenw do you know how we can exclude code from code coverage on Windows only?

hagenw: One way would be to create a `.coveragerc` on the fly in the runner on Windows and add the following content:

```config
[report]
exclude_lines =
    pragma: no cover
    pragma: no Windows cover
```

and change the line here to

```suggestion
        ):  # pragma: no Windows cover
```

.github/workflows/test.yml
```diff
@@ -43,7 +43,7 @@ jobs:
         ARTIFACTORY_USERNAME: ${{ secrets.ARTIFACTORY_USERNAME }}
         ARTIFACTORY_API_KEY: ${{ secrets.ARTIFACTORY_API_KEY }}
       run: |
-        python -m pytest
+        python -m pytest --cov-config=.coveragerc.$RUNNER_OS
```
frankenjoe, line 46: @hagenw I want the configuration file to be extended to `".coveragerc.[Windows,macOS,Linux]"`. Do you know the correct syntax to do that?

hagenw: An alternative is that you use an `if` statement, e.g.

```
    - name: Test with pytest on Windows
      env:
        ARTIFACTORY_USERNAME: ${{ secrets.ARTIFACTORY_USERNAME }}
        ARTIFACTORY_API_KEY: ${{ secrets.ARTIFACTORY_API_KEY }}
      run: |
        python -m pytest --cov-config=.coveragerc.windows
      if: matrix.os == 'windows-latest'
```

frankenjoe: Thanks. I now used `runner.os` and looks like it worked.

audbackend/core/backend.py
frankenjoe: Thanks, I now switched to custom configuration files for each OS and applied your suggestion.

hagenw:
For some backends we will loose a little bit of information as we no longer raise a `FileNotFoundError` when the file is really missing. But as some backends cannot differentiate between permission error and file not found errors, I don't see a better way in handling this then always raising a generic `BackendError`.

frankenjoe:
It also has the advantage that we don't have to update the package everytime they change the exception handling on the backend as e.g. in https://github.com/audeering/audbackend/pull/73

hagenw:
For `Backend.latest_version()` we write in the docstring:

image

If we change `versions()` to raise a `BackendError` we should also change `latest_version()` to do the same.
If we stay with the current behavior of `versions()` returning an empty list we should change maybe to
```
RuntimeError: if no version could be found
```

audbackend/core/backend.py
```diff
@@ -323,6 +338,7 @@ def ls(
             list of tuples (path, version)
 
         Raises:
+            BackendError: if an error is raised on the backend
             FileNotFoundError: if ``folder`` does not exist
```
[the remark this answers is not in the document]
frankenjoe: Right, I removed it.

frankenjoe:
> For `Backend.latest_version()` we write in the docstring:
> 
> image
> 
> If we change `versions()` to raise a `BackendError` we should also change `latest_version()` to do the same. If we stay with the current behavior of `versions()` returning an empty list we should change maybe to
> 
> ```
> RuntimeError: if no version could be found
> ```

I changed to:

image

audbackend/core/backend.py
```diff
@@ -399,6 +407,7 @@ def put_archive(
             verbose: show debug messages
 
         Raises:
+            BackendError: if an error is raised on the backend
             FileNotFoundError: if one or more files do not exist
```
hagenw, line 412: To highlight that we are talking about the argument for local files here:
```suggestion
            FileNotFoundError: if one or more ``files`` do not exist
```

audbackend/core/backend.py
```diff
@@ -399,7 +408,8 @@ def put_archive(
             verbose: show debug messages
 
         Raises:
-            FileNotFoundError: if one or more files do not exist
+            BackendError: if an error is raised on the backend
+            FileNotFoundError: if one or more ``files`` do not exist
             FileNotFoundError: if ``tmp_root`` does not exist
```
hagenw, line 413: Maybe I made a mistake here. I guess it should be ``src_root`` and not ``tmp_root``? At least there is no test for missing ``tmp_root``.

hagenw:
Tests for `BackendError` are missing for:

* `exists()`
* `latest_version()`
* `put_file()`
* `versions()`

audbackend/core/backend.py
frankenjoe: Would you prefer three single entries for `FileNotFoundError`:

```
            FileNotFoundError: if ``src_root`` does not exist
            FileNotFoundError: if one or more ``files`` do not exist
            FileNotFoundError: if ``tmp_root`` does not exist       
```

or should we combine them:

```
            FileNotFoundError: if ``src_root``, ``tmp_root``, or one or more ``files`` do not exist
```

frankenjoe:
> Tests for `BackendError` are missing for:
> 
>     * `exists()`
> 
>     * `latest_version()`
> 
>     * `put_file()`
> 
>     * `versions()`

Mhh, I guess it's not straightfoward to add tests for these functions. We probably need a custom `RaiseAlwaysBackend` or something like that. Not sure if this is really useful.

audbackend/core/backend.py
frankenjoe: For now I changed to:

image

hagenw: Your solution looks fine to me.

hagenw:
> > Tests for `BackendError` are missing for:
> > ```
> > * `exists()`
> > 
> > * `latest_version()`
> > 
> > * `put_file()`
> > 
> > * `versions()`
> > ```
> 
> Mhh, I guess it's not straightfoward to add tests for these functions. We probably need a custom `RaiseAlwaysBackend` or something like that. Not sure if this is really useful.

I think it also points to the problem that it depends on their implementation on the backend if the error will ever happen. E.g. `Artifactory.exists()` will always return `False` instead of raising a `BackendError`, see test in `tests/test_artifactory.py`. `versions()` and hence `latest_version()` might also never return a `BackendError`.

So, it's a tricky question how we handle this.
We could simply stick with the current description, but it is also somehow wrong when we state

image

but `Artifactory.exists()` returns `False` in this case.

frankenjoe:
> but `Artifactory.exists()` returns `False` in this case.

Oh yes, looks like we are catching some errors here, which we should no longer. So I suggest we change:

```python
    def _exists(
            self,
            path: str,
            version: str,
    ) -> bool:
        path = self._path(path, version)
        try:
            return audfactory.path(path).exists()
        except self._non_existing_path_error:
            return False
```

to:


```python
    def _exists(
            self,
            path: str,
            version: str,
    ) -> bool:
        path = self._path(path, version)
        return audfactory.path(path).exists()        
```

hagenw:
With `Backend.exists()` we also have a similar usability question as with `versions()`.
Do we really want that it raises backend errors, or should it better return `False`.

I also think we should not make a mixture as we currently do for `versions()` and `latest_version()`. If we want that `versions()` can return an empty list we should adjust its code that it can never raise a `BackendError`, and hence `latest_version()` can also never raise a `BackendError`

hagenw:
E.g.
```python
    def versions(
            self,
            path: str,
    ) -> typing.List[str]:
        r"""Versions of a file.

        Args:
            path: path to file on backend

        Returns:
            list of versions in ascending order

        Raises:
            ValueError: if ``path`` contains invalid character

        Examples:
            >>> backend.versions('name.ext')
            ['1.0.0', '2.0.0']

        """
        utils.check_path_for_allowed_chars(path)

        try:
            vs = self._versions(path)
        except Exception
            vs = []

        return audeer.sort_versions(vs)
```

frankenjoe:
> Yes, and then you can use the current `test_exists` in `tests/test_artifactory()` to check that a `BackendError` is raised.

Done- We now have tests for  `BackendError` for all functions.

frankenjoe:
> I also think we should not make a mixture as we currently do for `versions()` and `latest_version()`. If we want that `versions()` can return an empty list we should adjust its code that it can never raise a `BackendError`, and hence `latest_version()` can also never raise a `BackendError`

I would argue that even if `versions()` returns an empty list if not match is found, it can still raise a `BackendError`, e.g. if the repository does ot exist or a connection error occurs. Just like it is now implemented for `exists()`.

Issue #89, opened 2023-04-18
Should `backend.create()` create repository if it does not exist


frankenjoe:
Related to #86 and https://github.com/audeering/audbackend/issues/24

As discussed in #86 `backend.create()` should raise an error if the repository does not exist. The question is - should we allow or even encourage a backend to (if possible) create the repository if it does not exist? The advantage I see is that  a user does not have to manually create the repository beforehand. The disadvantage might be that a user accidentally creates a new repository when she actually wanted to work from an existing one.

frankenjoe:
I think generally it could be handy if a backend can create a repository, but maybe we should introduce an argument like `audbackend.create(..., allow_creation=True)` which by default we set to `False`?

frankenjoe:
> Again, we have the problem here that it my lead to different behaviors on different backends, as there might be backends where the user never has permissions to create a repository.

That's ok, in that case the backend should raise an error and the user has to manually create the repository. It's only important that - if `create()` does not raise an error - the repository exists.

Issue #24, opened 2021-07-23
What should backend.exists() return for non-existing backends?


hagenw:
It might be that you specify a backend for which the host does not exists.

For `audbackend.FileSystem`, `backend.exists(path)` will then return `False` for every `path` variable,
but for `audbackend.Artifactory` it will fail and raise an `requests.exceptions.ConnectionError` error.

Which of those is the expected behavior when a host does not exists?

frankenjoe:
Ideally we already raise an error when you try to create a backend with an invalid host?

hagenw:
> Ideally we already raise an error when the user tries to create a backend with an invalid host?

Yes, I think that would be a good solution. For `FileSystem` you could even create the `host` on the fly (and maybe we even do it in some of our tests), but I would argue that it is more important that the different backends behave in the same way.

frankenjoe:
>  but I would argue that it is more important that the different backends behave in the same way.

agree

frankenjoe:
We now ensure that the repository must exist, see #91 otherwise an error is raised already when `audbackend.create()` is called.

PR #91, opened 2023-04-18
Create repository if it does not exist

frankenjoe:
Relates to #89, https://github.com/audeering/audbackend/issues/86, #24 

When a backend is created it will create the repository if it does not exist. This may raise a `BackendError` and the docstring has been extended accordingly:

image

image

image

Other changes:

* It is no longer possible to use a folder inside an Artifactory repository as repository
* To avoid a BackendError is now raised in the docstring examples, the Artifactory backend has been removed. 
* The docstring examples have been further simplified by directly setting the working directory to the temporary folder.
* The test for listing an empty repository is re-enabled.

### DIFF
--- a/audbackend/core/api.py
+++ b/audbackend/core/api.py
@@ -30,8 +30,8 @@ def available() -> typing.Dict[str, typing.List[Backend]]:
     Examples:
         >>> list(available())
         ['artifactory', 'file-system']
-        >>> available()['artifactory']
-        [('audbackend.core.artifactory.Artifactory', 'https://host.com', 'repo')]
+        >>> available()['file-system']
+        [('audbackend.core.filesystem.FileSystem', 'host', 'doctest')]
 
     """  # noqa: E501
     result = {}
@@ -66,11 +66,11 @@ def create(
 
     Examples:
         >>> create(
-        ...     'artifactory',
-        ...     'https://host.com',
+        ...     'file-system',
+        ...     'host',
         ...     'repo',
         ... )
-        ('audbackend.core.artifactory.Artifactory', 'https://host.com', 'repo')
+        ('audbackend.core.filesystem.FileSystem', 'host', 'repo')
 
     """
     if name not in backend_registry:
--- a/audbackend/core/api.py
+++ b/audbackend/core/api.py
@@ -3,6 +3,7 @@ import typing
 from audbackend.core.artifactory import Artifactory
 from audbackend.core.backend import Backend
 from audbackend.core.filesystem import FileSystem
+from audbackend.core import utils
 
 
 backends = {}
@@ -62,6 +63,7 @@ def create(
         backend object
 
     Raises:
+        BackendError: if an error is raised on the backend
         ValueError: if registry name does not exist
 
     Examples:
@@ -87,7 +89,11 @@ def create(
         backends[name][host] = {}
     if repository not in backends[name][host]:
         backend = backend_registry[name]
-        backends[name][host][repository] = backend(host, repository)
+        backends[name][host][repository] = utils.call_function_on_backend(
+            backend,
+            host,
+            repository,
+        )
     return backends[name][host][repository]
 
 
--- a/audbackend/core/api.py
+++ b/audbackend/core/api.py
@@ -68,8 +68,8 @@ def create(
     :class:`audbackend.BackendError`
     is raised.
 
-    Use :func:`audbackend.avilable`
-    to list available backend alias.
+    Use :func:`audbackend.available`
+    to list available backend aliases.
 
     Args:
         name: alias under which backend class is registered
--- a/audbackend/core/api.py
+++ b/audbackend/core/api.py
@@ -20,7 +20,7 @@ def available() -> typing.Dict[str, typing.List[Backend]]:
     r"""List available backends.
 
     Returns a dictionary with
-    registered backend names as keys
+    registered backend aliases as keys
     (see :func:`audbackend.register`)
     and a list with backend instances as values
     (see :func:`audbackend.create`).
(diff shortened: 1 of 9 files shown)


audbackend/core/artifactory.py
```diff
@@ -26,6 +24,11 @@ def __init__(
     ):
         super().__init__(host, repository)
 
+        self.root = f'{self.host}/{self.repository}'
+        path = audfactory.path(self.root)
+        if not path.exists():
+            audfactory.path(self.root).mkdir()
```
hagenw, line 30: This does not create a repository, this creates a folder in an existing repository.

hagenw: I think code to create repositories looks more like the examples at https://github.com/devopshq/artifactory#repositorylocal, even though I don't understand them yet.

frankenjoe: Ah great, I was already wondering how to create a "real" repository.

What is a bit problematic at the moment is that in our tests we actually use a folder on a repo as repo. But I guess that is anyway bad practice and we should change it once we are able to properly create a repository inside the backend. Let's see...

frankenjoe: I guess we should then forbid to use a folder in a repo as repo, right?

hagenw:
Would be nice to expand the docstring of `create()` along those lines:

* mention in the description that it tries to create non-existing repos on the backend.
* "name: backend registry name", what's that and where can a user find available values (or maybe list available values directly here as this will not change very often)
* "repository". At some point we need to stay which kind of repositories we are support/recommend on Artifactory as I don't know if the backend will work with all types of repositories you can create on Artifactory. But maybe it is better to add this discussion to the description part of the docstring

audbackend/core/artifactory.py
hagenw: In general it doesn't seem to be a problem as it works. But it is not a repository, as a repository has special features on Artifactory (e.g. it can be mirrored and stuff like this). So yes, we should forbid to use a folder as repository on the Artifactory backend.

audbackend/core/api.py
```diff
@@ -62,15 +63,16 @@ def create(
         backend object
 
     Raises:
+        BackendError: if an error is raised on the backend
```
hagenw, line 66: This sounds a little bit too generic to me.
Maybe change to
```suggestion
        BackendError: if an error is raised on the backend
            when trying to access or create the repository
```
Or are there other interactions which could lead to this error?

frankenjoe: I guess so, e.g. a connection error. So maybe we leave it as is but add more explanation what might get wrong in the description as suggested in https://github.com/audeering/audbackend/pull/91#issuecomment-1514293952

hagenw: OK, makes sense.

audbackend/core/artifactory.py
frankenjoe: I have now changed the code to create a repository instead of a folder inside an existing one. Actually it is no longer possible to initialize an Artifactory backend with a folder and a (failing) test has been added.

frankenjoe:
> Would be nice to expand the docstring of `create()` along those lines:

Please have a look at the screenshots in the description.

tests/test_artifactory.py
```diff
@@ -4,41 +4,46 @@
...
+def backend(request):
 
+    backend = audbackend.Artifactory(
+        pytest.HOSTS['artifactory'],
+        f'unittest-{audeer.uid()[:8]}',
+    )
+
+    yield backend
+
+    # TODO: replace with audbackend.delete() when available
```
frankenjoe, line 17: It would be handy to introduce such a function. But it's also somehow dangerous of course :)

hagenw: It depends, you could argue that `audbackend` is for developers anyway.

audbackend/core/conftest.py
```diff
@@ -1,29 +1,36 @@
...
+
 import pytest
 
 import audbackend
 import audeer
 
 
-@pytest.fixture(autouse=True)
-def create_backend(doctest_namespace):
+@pytest.fixture(scope='function', autouse=True)
```
hagenw, line 10: Is scope `function` correct, or can this be reused across functions?

tests/test_backend.py
```diff
@@ -134,27 +140,55 @@ def test_archive(tmpdir, files, name, folder, version, tmp_root, backend):
...
-        (
-            'artifactory',
-            audbackend.Artifactory,
-        ),
+        # (
+        #     'artifactory',
+        #     pytest.HOSTS['artifactory'],
+        #     f'unittest-{audeer.uid()[:8]}',
+        #     audbackend.Artifactory,
+        # ),
```
hagenw, line 156: Why is this commented?

audbackend/core/artifactory.py
hagenw: I cannot find any leftover repositories on Artifactory, so it seems to work:

image

audbackend/core/conftest.py
frankenjoe: I think it makes writing the tests easier if we always start from a fresh repository.

tests/test_backend.py
frankenjoe: ups, I removed the comments again

audbackend/core/conftest.py
hagenw: If it does not slow down the tests considerable, it seems indeed the safer option.
Have you any experience how long it takes to create the repo?

frankenjoe: ~1-2 seconds

hagenw: OK, not ideal but as it is used for the docstring tests where we shouldn't expect a big increase in tests it should be fine.

frankenjoe: Oh wait a moment. You were actually referring to the docstring examples here, which is using a file-system backend only. This has almost no overhead at all.

Issue #95, opened 2023-04-19
Rename folder to sub-path


hagenw:
As @frankenjoe has correctly stated somewhere, it might happen that we have backends that do not have a concept of a folder. So it might be better to avoid the wording `folder` and use something like `sub-path` instead.

hagenw:
Closed by https://github.com/audeering/audbackend/pull/96

Issue #57, opened 2022-12-08
Ensure backend paths start with `/`


frankenjoe:
In #53 we define that the root of a backend as `/`. Accordingly we should ensure that a backend path always starts with `/`. For backward compatibility I suggest we do this in `utils.check_path_and_ext()`.

hagenw:
Mh, maybe I was to strong requesting `/` for the root of the backend, because you cannot ask for relative paths anyway, can you?
So there should be no difference between `/folder/file.ext` and `path/file.ext`.

On the other you are also right, starting always with `/` makes it more obvious that we are dealing with the root of the backend here.

frankenjoe:
Yes, I suggest we accept both (`/folder/file.ext` and `folder/file.ext`), but internally always convert to `/folder/file.ext`.

hagenw:
OK

frankenjoe:
Since we will anyway introduce some breaking changes I would propose that we now enforce that a backend path has to start with `'/'`. So far we have no use-case for relative paths, but maybe someday and then it's easier to implement when we now ensure all paths are passed as absolute paths.

frankenjoe:
@hagenw before I start working on a PR, do you agree with my suggestion to enforce `'/'`?

hagenw:
Yes, I think it makes indeed more future proof to require `/` when referring to an absolute path, even though we don't have any method that supports a relative path yet.

PR #96, opened 2023-04-20
Backend.ls(): support full path

### LINKED ISSUE #59: Support backend.ls('folder/name.ext') for single files
`audbackend.ls()` as added in #53 does not support listing of single files as this would require to add `ext` as an argument.

As discussed in #58 we are going to remove the `ext` argument altogether. Afterwards we should update `ls()` to support also listing of single files.

frankenjoe:
Closes #59 #95 

image

Additional changes:

* `Backend.versions()` is now implemented based on `Backend.ls()` and `Backend._versions()` is removed
* The use of the word `folder` is avoided and instead we talk about `sub-path`

### DIFF
--- a/audbackend/core/backend.py
+++ b/audbackend/core/backend.py
@@ -332,16 +332,16 @@ class Backend:
         with path and version.
         It is possible to provide a sub-path
         e.g. ``/sub/path/``.
-        In that case all files 
-        that start with the sub-path are returned.         
+        In that case all files
+        that start with the sub-path are returned.
         When ``path`` is set to ``'/'``
         a (possibly empty) list with
-        all files on the backend is returned.        
+        all files on the backend is returned.
 
         Args:
             path: (sub-)path on backend.
                 If a sub-path is provided,
-                it has to end on ``/`` 
+                it has to end on ``/``
             latest_version: if multiple versions of a file exist,
                 only include the latest
             pattern: if not ``None``,
--- a/audbackend/core/backend.py
+++ b/audbackend/core/backend.py
@@ -310,32 +310,38 @@ class Backend:
             self,
             folder: str,
     ) -> typing.List[typing.Tuple[str, str, str]]:  # pragma: no cover
-        r"""List all files under folder.
+        r"""List all files under (sub-)path.
 
-        If folder does not exist an error should be raised.
+        * If path does not exist an error should be raised.
+        * If path ends on `/` it is a sub-path
 
         """
         raise NotImplementedError()
 
     def ls(
             self,
-            folder: str = '/',
+            path: str = '/',
             *,
             latest_version: bool = False,
             pattern: str = None,
             suppress_backend_errors: bool = False,
     ) -> typing.List[typing.Tuple[str, str]]:
-        r"""List all files under folder.
+        r"""List files on backend.
 
         Returns a sorted list of tuples
         with path and version.
-        When ``folder`` is set to the
-        root of the backend (``'/'``)
+        It is possible to provide a sub-path
+        e.g. ``/sub/path/``.
+        In that case all files 
+        that start with the sub-path are returned.         
+        When ``path`` is set to ``'/'``
         a (possibly empty) list with
-        all files on the backend is returned.
+        all files on the backend is returned.        
 
         Args:
-            folder: folder on backend
+            path: (sub-)path on backend.
+                If a sub-path is provided,
+                it has to end on ``/`` 
             latest_version: if multiple versions of a file exist,
                 only include the latest
             pattern: if not ``None``,
@@ -351,8 +357,8 @@ class Backend:
         Raises:
             BackendError: if ``suppress_backend_errors`` is ``False``
                 and an error is raised on the backend,
-                e.g. ``folder`` does not exist
-            ValueError: if ``folder`` contains invalid character
+                e.g. ``path`` does not exist
+            ValueError: if ``path`` contains invalid character
 
         Examples:
             >>> backend.ls()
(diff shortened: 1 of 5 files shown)


audbackend/core/backend.py
```diff
@@ -652,13 +663,6 @@ def versions(
...
-            self._versions,
-            path,
-            suppress_backend_errors=suppress_backend_errors,
-            fallback_return_value=[],
-        )
-
-        return audeer.sort_versions(vs)
+        paths = self.ls(path, suppress_backend_errors=suppress_backend_errors)
+        vs = [v for _, v in paths]
+        return vs
```
hagenw, line 668: As we now use `self.ls()` you should delete the definition of `_versions()` done a few lines before this method.

audbackend/core/filesystem.py
```diff
@@ -75,32 +89,46 @@ def _get_file(
...
     ):
-        r"""List all files under folder."""
+        r"""List all files under (sub-)path."""
+
+        if path.endswith('/'):  # find files under sub-path
+
+            path = self._expand(path)
+            if not os.path.exists(path):
+                utils.raise_file_not_found_error(path)
+            paths = audeer.list_file_names(path, recursive=True)
```
hagenw, line 101: I think we should also allow for hidden file names (and should add a test for them):

```suggestion
            paths = audeer.list_file_names(path, recursive=True, hidden=True)
```

audbackend/core/filesystem.py
```diff
@@ -75,32 +89,46 @@ def _get_file(
...
-        folder = self._folder(folder)
+        else:  # find versions of path
 
-        if not os.path.exists(folder):
-            utils.raise_file_not_found_error(folder)
+            root, _ = self.split(path)
+            root = self._expand(root)
+            vs = audeer.list_dir_names(
+                root,
+                basenames=True,
```
hagenw, line 109: And here the same

```suggestion
                basenames=True,
                hidden=True,
```

tests/test_backend.py
```diff
@@ -487,18 +495,22 @@ def test_ls(tmpdir, backend):
...
+        ('/', False, '*.foo', [('file.foo', '1.0.0')] + sub_content),
+        ('/', True, '*.foo', [('file.foo', '1.0.0')] + sub_content_latest),
+        ('sub/', False, None, sub_content),
+        ('sub/', True, None, sub_content_latest),
+        ('sub/', False, '*.bar', []),
+        ('sub/', True, '*.bar', []),
+        ('file.bar', False, None, content_bar),
+        ('file.bar', True, None, [('file.bar', '2.0.0')]),
+        ('sub/file.foo', False, None, sub_content),
+        ('sub/file.foo', True, None, sub_content_latest),
```
hagenw, line 510:
```suggestion
        ('sub/', False, None, sub_content),
        ('sub/', True, None, sub_content_latest),
        ('sub/', False, '*.bar', []),
        ('sub/', True, '*.bar', []),
        ('file.bar', False, None, content_bar),
        ('file.bar', True, None, [('file.bar', '2.0.0')]),
        ('sub/file.foo', False, None, sub_content),
        ('sub/file.foo', True, None, sub_content_latest),
        ('/sub/', False, None, sub_content),
        ('/sub/', True, None, sub_content_latest),
        ('/sub/', False, '*.bar', []),
        ('/sub/', True, '*.bar', []),
        ('/file.bar', False, None, content_bar),
        ('/file.bar', True, None, [('file.bar', '2.0.0')]),
        ('/sub/file.foo', False, None, sub_content),
        ('/sub/file.foo', True, None, sub_content_latest),
```

audbackend/core/backend.py
```diff
@@ -308,34 +308,45 @@ def latest_version(
...
     ) -> typing.List[typing.Tuple[str, str]]:
-        r"""List all files under folder.
+        r"""List files on backend.
 
         Returns a sorted list of tuples
         with path and version.
-        When ``folder`` is set to the
-        root of the backend (``'/'``)
+        If a full path
+        (e.g. ``/sub/file.ext``)
```
hagenw, line 334: Maybe change to 

```suggestion
        (e.g. ``sub/file.ext``)
```
as ``/`` is not really needed.

audbackend/core/backend.py
```diff
@@ -308,34 +308,45 @@ def latest_version(
...
+        is provided,
+        all versions of the path are returned.
+        If a sub-path
+        (e.g. ``/sub/``)
```
hagenw, line 338: Here the same

```suggestion
        (e.g. ``sub/``)
```

audbackend/core/backend.py
frankenjoe: Thanks, I forgot to remove it. Done.

audbackend/core/backend.py
frankenjoe: You are right, at the moment it's not needed. So I will apply your suggestion for now.

However, I think we should enforce that a backend path has to start with `/`. Maybe someday there will be a use-case for using relative backend paths, so it's better to now insist on absolute paths. But I will tackle this once when we solve https://github.com/audeering/audbackend/issues/57.

audbackend/core/filesystem.py
frankenjoe: Here `True` is the default value, but it doesn't hurt to be verbose of course. Maybe we'll change it someday in `audeer`.

audbackend/core/filesystem.py
frankenjoe: I added tests for hidden files.

tests/test_backend.py
```diff
@@ -461,23 +461,31 @@ def test_ls(tmpdir, backend):
...
     ]
-    content_latest = content[-1:] + sub_content_latest
-    content += sub_content
+    content_latest = [
+        ('file.bar', '2.0.0'),
+        ('file.foo', '1.0.0'),
+    ]
+    sub_content = [
+        ('sub/.file.foo', '1.0.0'),
+        ('sub/.file.foo', '2.0.0'),
```
hagenw, line 479: Could you also add a test for a hidden folder as different functions might be used in the backend to discover them (as it is the case in the current file-system backend).

frankenjoe: Good idea. I extended the tests accordingly.

PR #98, opened 2023-04-20
Add audbackend.delete()

### LINKED ISSUE #92: Add function to delete repository
It might be useful to have `audbackend.delete()` to delete a repository.

frankenjoe:
Closes #92 

This adds `audbackend.delete()` as counterpart to `audbackend.create()`, i.e. it deletes the respository and removes the instance from the available backend instances. Since it is not directly available as function of `Backend` it avoids the problem that a user does, e.g.:

```
backend = audbackend.create(...)
backend.delete()
backend.ls()  # would fail since repository was deleted
```

image

Other changes:

* The docstring of `audbackend.available()` and `audbackend.create()` are slightly improved
* All function so the API are not tested in `tests/test_api.py`
* The fixture that generates the backend objects is moved to `tests/conftest.py` and shared by all test files

### DIFF
--- a/audbackend/core/api.py
+++ b/audbackend/core/api.py
@@ -141,22 +141,10 @@ def delete(
             has been registered
 
     Examples:
-        >>> create(
-        ...     'file-system',
-        ...     'host',
-        ...     'doctest',
-        ... ).ls()
+        >>> create('file-system', 'host', 'doctest').ls()
         [('a.zip', '1.0.0'), ('a/b.ext', '1.0.0'), ('name.ext', '1.0.0'), ('name.ext', '2.0.0')]
-        >>> delete(
-        ...     'file-system',
-        ...     'host',
-        ...     'doctest',
-        ... )
-        >>> create(
-        ...     'file-system',
-        ...     'host',
-        ...     'doctest',
-        ... ).ls()
+        >>> delete('file-system', 'host', 'doctest')
+        >>> create('file-system', 'host', 'doctest').ls()
         []
 
     """  # noqa: E501
--- a/audbackend/core/api.py
+++ b/audbackend/core/api.py
@@ -87,11 +87,7 @@ def create(
             has been registered
 
     Examples:
-        >>> create(
-        ...     'file-system',
-        ...     'host',
-        ...     'doctest',
-        ... )
+        >>> create('file-system', 'host', 'doctest')
         ('audbackend.core.filesystem.FileSystem', 'host', 'doctest')
 
     """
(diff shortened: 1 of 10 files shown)


audbackend/core/api.py
```diff
@@ -115,6 +117,54 @@ def create(
...
+        ...     'file-system',
+        ...     'host',
+        ...     'doctest',
+        ... ).ls()
+        [('a.zip', '1.0.0'), ('a/b.ext', '1.0.0'), ('name.ext', '1.0.0'), ('name.ext', '2.0.0')]
+        >>> delete(
+        ...     'file-system',
+        ...     'host',
+        ...     'doctest',
+        ... )
```
hagenw, line 154: This feels slightly unsymmetrical. `create()` does not (always) create a new repo, but just an instance to access the repo, whereas `delete()` really deletes the repo.
But I also don't have an idea how to improve this.

audbackend/core/api.py
```diff
@@ -115,6 +117,54 @@ def create(
...
+        >>> create(
+        ...     'file-system',
+        ...     'host',
+        ...     'doctest',
+        ... ).ls()
+        []
```
hagenw, line 160: I would make this a little bit more compact:

```suggestion
        >>> create('file-system', 'host', 'doctest').ls()
        [('a.zip', '1.0.0'), ('a/b.ext', '1.0.0'), ('name.ext', '1.0.0'), ('name.ext', '2.0.0')]
        >>> delete('file-system', 'host', 'doctest')
        >>> create('file-system', 'host', 'doctest').ls()
        []
```

or maybe

```suggestion
        >>> backend = create('file-system', 'host', 'doctest')
        >>> backend.ls()
        [('a.zip', '1.0.0'), ('a/b.ext', '1.0.0'), ('name.ext', '1.0.0'), ('name.ext', '2.0.0')]
        >>> delete('file-system', 'host', 'doctest')
        >>> backend = create('file-system', 'host', 'doctest')
        >>> backend.ls()
        []
```

to tackle my other comment to this docstring.

audbackend/core/api.py
frankenjoe: Right, `create()` only creates a new repo if it does not exist yet. Maybe `audbackend.open()` would be a better fit?

frankenjoe: or maybe `init()`?

hagenw: I created https://github.com/audeering/audbackend/issues/99 to handle this outside of this pull request.

hagenw:
Can you update the screenshot in the description, please.

PR #100, opened 2023-04-21
Split into audbackend.access() and audbackend.create()

### LINKED ISSUE #99: Consider renaming audbackend.create()
As discussed in https://github.com/audeering/audbackend/pull/98#discussion_r1172625244 where we are adding `audbackend.delete()`, `audbackend.create()` is slightly misleading as in most cases it does not create a new repository on the provided backend, but just returns a `Backend` instance that has access to that repo. If the repo does not exists it is also created, but this is just a minor use-case.

A better fit would be to call it `audbackend.open()` or `audbackend.init()`.

frankenjoe:
Closes #99 

image

image

This is complemented by:

image

### DIFF
--- a/audbackend/core/api.py
+++ b/audbackend/core/api.py
@@ -16,6 +16,72 @@ backend_registry = {
 r"""Backend registry."""
 
 
+def access(
+        name: str,
+        host: str,
+        repository: str,
+) -> Backend:
+    r"""Access repository.
+
+    Returns a backend instance
+    for the ``repository``
+    on the ``host``.
+    The instance is an object of the class
+    registered under the alias ``name``
+    with :func:`audbackend.register`.
+
+    If the repository does not exist
+    or cannot be accessed for other reasons
+    (e.g. user lacks permission)
+    an error of type
+    :class:`audbackend.BackendError`
+    is raised.
+
+    Use :func:`audbackend.available`
+    to list available backend instances.
+
+    Args:
+        name: alias under which backend class is registered
+        host: host address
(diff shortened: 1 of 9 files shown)


audbackend/core/api.py
```diff
@@ -82,43 +157,27 @@ def create(
...
+    return backend
 
 
 def delete(
         name: str,
         host: str,
         repository: str,
 ):
-    r"""Delete repository and remove backend instance.
+    r"""Delete repository.
```
hagenw, line 180: What did "remove backend instance" referred to?

audbackend/core/api.py
```diff
@@ -82,43 +157,27 @@ def create(
         backend object
 
     Raises:
-        BackendError: if an error is raised on the backend
+        BackendError: if an error is raised on the backend,
+            e.g. repository cannot be created
```
hagenw, line 161: If I saw it correctly inside the `Artifactory._create()` implementation it will also raise this error if the repo exists already, correct?
Then I think we should also mention this here explicitly:

```suggestion
            e.g. repository exists already
            or cannot be created
```

audbackend/core/api.py
frankenjoe: It's explained in the docstring, so I don't think we need to mention it here.

frankenjoe: image

The caption is now aligned with that of the other functions.

hagenw: Thanks.

PR #103, opened 2023-04-24
TST: add cleanup of leftover Artifactory repos

### LINKED ISSUE #97: TST: deleting temporary repositories can fail
During the tests we create temporary repositories on https://audeering.jfrog.io/.

Unfortunately, it can sometimes happen that an error occurs when the repositories are deleted:

image

If this occurs more often we could try to catch the error and do a couple of retries before we actually raise.

For full output, see https://github.com/audeering/audbackend/actions/runs/4752902382/jobs/8443828758

hagenw:
Closes #97 

This introduces two steps to try to remove repositories on Artifactory

* It tries up to 3 times to remove the repo and catches `BackendErrors` in the `backend()` fixture. After that it raises a `RuntimeError` and asks the user to manually delete the repo.
* In addition, it adds a session UID to the repo name, so it is easier to recognize remaining repositories on the host

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -74,7 +74,7 @@ def backend(request):
     r"""Create and delete a repository on the backend."""
     name = request.param
     host = pytest.HOSTS[name]
-    repository = f'unittest-{audeer.uid()[:8]}'
+    repository = f'unittest-{pytest.UID}-{audeer.uid()[:8]}'
 
     backend = audbackend.create(name, host, repository)
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -38,6 +38,10 @@ def cleanup_artifactory():
     (https://github.com/audeering/audbackend/issues/97),
     we try to clean up the host
     everytime we run the tests.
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -20,19 +22,78 @@
...
+                    f'Cleaning up of repo {repo} failed.\n'
+                    'Please delete remaining repositories manually with:\n'
+                )
+                for remaining_repo in remaining_repos:
+                    error_msg += (
+                        f"'audbackend.delete({name}, {host}, "
+                        f"{remaining_repo})',\n"
+                    )
+                error_msg += f'The original error message was {ex}.'
+                raise RuntimeError(error_msg)
```
frankenjoe, line 76: So the idea is that this let's the CI job fail and then the developer will look up which repositories were not deleted and do it manually. I wonder if it's simpler we raise the error directly in the backend fixture instead. Then it's always only one repository the developer has to delete and we don't need an additional `cleanup_artifactory()` fixture.

PR #104, opened 2023-04-24
Enforce backend path starts with /

### LINKED ISSUE #57: Ensure backend paths start with `/`

frankenjoe:
Closes #57 

* An error is raised if backend path does not start with '/'
* Successive '/' are automatically joined to one

### Examples

image

image

### DIFF
--- a/audbackend/core/backend.py
+++ b/audbackend/core/backend.py
@@ -68,6 +68,7 @@ class Backend:
             BackendError: if an error is raised on the backend,
                 e.g. ``path`` does not exist
             ValueError: if ``path`` contains invalid character
+                or does not start with ``/``
 
         Examples:
             >>> backend.checksum('/name.ext', '1.0.0')
@@ -130,6 +131,7 @@ class Backend:
                 and an error is raised on the backend,
                 e.g. ``path`` does not exist
             ValueError: if ``path`` contains invalid character
+                or does not start with ``/``
 
         Examples:
             >>> backend.exists('/name.ext', '1.0.0')
@@ -180,6 +182,7 @@ class Backend:
             RuntimeError: if extension of ``src_path`` is not supported
             RuntimeError: if ``src_path`` is a malformed archive
             ValueError: if ``src_path`` contains invalid character
+                or does not start with ``/``
 
         Examples:
             >>> backend.get_archive('/a.zip', '.', '1.0.0')
@@ -243,6 +246,7 @@ class Backend:
             PermissionError: if the user lacks write permissions
                 for ``dst_path``
             ValueError: if ``src_path`` contains invalid character
+                or does not start with ``/``
 
         Examples:
             >>> os.path.exists('dst.pth')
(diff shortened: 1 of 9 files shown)


audbackend/core/backend.py
```diff
@@ -290,19 +294,26 @@ def join(
             path joined by :attr:`Backend.sep`
 
         Raises:
-            ValueError: if joined path contains invalid character
+            ValueError: if ``path`` contains invalid character
+                or does not start with ``'/'``,
```
hagenw, line 298: As `join()` does not communicate with the backend it feels a little bit strange that you also have to start the first path with `/`. Of cause if you don't do it you cannot use the path with any other function, but still there shouldn't be a reason why you have to start the path with `'/'`.

audbackend/core/backend.py
```diff
@@ -637,13 +653,14 @@ def split(
 
         Raises:
             ValueError: if ``path`` contains invalid character
+                or does not start with ``'/'``
```
hagenw, line 656: Here it is similar to `join()`. There is no need that the path has to start with `'/'` as we don't communicate with the backend.

At the moment the docstring reads:
> Split path on backend into root and basename.

I guess the better wording might be: "Split path on backend into sub-path and basename."

BTW, what is the expected outcome of
```python
split('/f.ext')
```
Is it
```
('/', 'f.ext')
```
or
```
('/f.ext')
```
?
Maybe add a few explicit tests to `tests/test_backend.py` as `split()` is only indirectly tested at the moment.

audbackend/core/backend.py
frankenjoe: Yes, I was struggling with this decision. But as you said, at the moment there is not use-case for joining with a path that does not start with `/`, so I thought it doesn't hurt to be strict now and maybe soften the constraints at later a point that the other way round.

hagenw: Yes, just wanted to check if you had thought about it ;)

Your solution makes also sense as it is consistent, there is no usage of a path without `/`, and we can remove that restriction when introducing functionality with relative paths.
So let's stay with your decision.

audbackend/core/backend.py
hagenw: As discussed in https://github.com/audeering/audbackend/pull/104#discussion_r1176073420 we stay with raising the error when not starting with `/`

frankenjoe: Very good point!

As we output a tuple, I think it should be `('/', 'f.ext')`, then `join(split('/f.ext'))` will also work.

And I think the sub-path should always end on `/`, i.e.:

```
backend.split('/sub/f.ext')
```
```
('/sub/', 'f.ext')
```

hagenw: I agree regarding always returning a tuple, this is also in line with `os.path.split`:
```python
>>> os.path.split('/f.ext')
('/', 'f.ext')
```
I'm not so sure about always ending the sub-path on `/` as it is redundant information and not the behavior of `os.path.split`:
```python
>>> os.path.split('/a/f.ext')
('/a', 'f.ext')
```

frankenjoe: I think it makes sense as with `ls()` we also require that a sub-path ends on `/`. And otherwise, the following would not work:

```python
>>> backend.ls(backend.split('/sub/f.ext')[0])
```

There is also a fundamental difference between the path on a file-system and the path on a backend. On a file-system we cannot have a folder and file with the same name in the same directory. But with a backend path we don't have this limitation, e.g. you can do:

```python
>>> backend.put_file(file, '/a', '1.0.0')
>>> backend.put_file(file, '/a/a', '1.0.0')
```

So I think it is important we always distinguish between `/a` and `/a/`.

frankenjoe: And maybe we should then also raise an error with:

```
>>> backend.join('/sub', 'f.ext')
```

frankenjoe: But maybe it's fine like this, since the output of the function will be a valid backend path.

hagenw: Yes, I would still allow `backend.join('/sub', 'f.ext')` as you also allow for `backend.join('//sub//', 'f.ext')`

frankenjoe: I added tests for `split()` and also extended the docstring example to:

image

hagenw: Great, thanks

Issue #90, opened 2023-04-18
Artifactory: store authentication


frankenjoe:
Relates to #50 

In the Artifactory backend we currently rely on the `path` function from `audfactory`. Every time it is called it tries to read username and API key. If we assume that those will not change during a session, we can maybe retrieve them once when the backend object is created and reuse them afterwards.

hagenw:
Good idea, this might also allow us to tackle https://github.com/audeering/audbackend/issues/50, as the authentication is the biggest feature of `audfactory` we are using at the moment.

frankenjoe:
Done in #50

PR #106, opened 2023-04-26
Remove dependency to audfactory

### LINKED ISSUE #50: Remove dependency on audfactory
As we duplicate lots of functionality from `audfactory` here and the package is not used in other packages, it would make sense to completely add the code here and deprecate `audfactory`.

frankenjoe:
Closes #50 #90 

* The functions to authenticate and download / deploy an artifact are extracted from `audfactory` and implemented in the `Artifactory` class
* Authentication done only once when the backend is created and re-used afterwards
* The dependency to `audfactory` is removed and a dependency to `dohq-artifactory` is set instead

### DIFF
--- a/audbackend/core/artifactory.py
+++ b/audbackend/core/artifactory.py
@@ -78,6 +78,42 @@ def _authentication(host) -> typing.Tuple[str, str]:
     return username, apikey
 
 
+def _deploy(
+        src_path: str,
+        dst_path: artifactory.ArtifactoryPath,
+        *,
+        verbose: bool = False,
+):
+    r"""Deploy local file as an artifact.
+
+    Args:
+        src_path: local file path
+        dst_path: path on Artifactory
+        verbose: show information on the upload process
+
+    """
+    if verbose:  # pragma: no cover
+        desc = audeer.format_display_message(
+            f'Deploy {src_path}',
+            pbar=False,
+        )
+        print(desc, end='\r')
+
+    md5 = utils.md5(src_path)
+
+    if not dst_path.parent.exists():
+        dst_path.parent.mkdir()
+    with open(src_path, "rb") as fobj:
+        dst_path.deploy(
+            fobj,
+            md5=md5,
+        )
+
+    if verbose:  # pragma: no cover
+        # Final clearing of progress line
+        print(audeer.format_display_message(' ', pbar=False), end='\r')
+
+
 def _download(
         src_path: artifactory.ArtifactoryPath,
         dst_path: str,
@@ -88,17 +124,10 @@ def _download(
     r"""Download an artifact.
 
     Args:
-        src_path: artifact URL
-        dst_path: path to store the artifact
+        src_path: local file path
+        dst_path: path on Artifactory
         chunk: amount of data read at once during the download
         verbose: show information on the download process
-
-    Returns:
-        path to local artifact
-
-    Raises:
-        RuntimeError: if artifact cannot be found,
-            or you don't have access rights to the artifact
 
     """
     src_size = artifactory.ArtifactoryPath.stat(src_path).size
@@ -127,8 +156,6 @@ def _download(
             if os.path.exists(dst_path):
                 os.remove(dst_path)  # pragma: no cover
             raise
-
-    return dst_path
 
 
 class Artifactory(Backend):
@@ -335,7 +362,7 @@ class Artifactory(Backend):
     ):
         r"""Put file to backend."""
         dst_path = self._path(dst_path, version)
-        audfactory.deploy(src_path, str(dst_path), verbose=verbose)
+        _deploy(src_path, dst_path, verbose=verbose)
 
     def _remove_file(
             self,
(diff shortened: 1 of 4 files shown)


audbackend/core/artifactory.py
```diff
@@ -1,13 +1,159 @@
...
+            pbar=False,
+        )
+        print(desc, end='\r')
+
+    md5 = utils.md5(src_path)
+
+    if not dst_path.parent.exists():
+        dst_path.parent.mkdir()
+    with open(src_path, 'rb') as fd:
+        dst_path.deploy(fd, md5=md5)
```
frankenjoe, line 106: In the function from `audfactory` also `sha1` and `sha256` were calculated and passed to the function. But in `audbackend` we only rely on `md5`, so I am not sure if we want this overhead?

audbackend/core/artifactory.py
```diff
@@ -1,13 +1,143 @@
...
+        host: host address
+
+    Returns:
+        username and API key
+
+    """
+    username = os.getenv('ARTIFACTORY_USERNAME', None)
+    apikey = os.getenv('ARTIFACTORY_API_KEY', None)
+
+    if apikey is None or username is None:  # pragma: no cover
```
frankenjoe, line 53: When running the test locally it reads from config file, but in the CI job we rely on environment variables. So I guess it's ok to exclude this from coverage. Otherwise, we would somehow need to overwrite the `global_config`  variable file in the `artifactory` package. I guess it's possible, but maybe not worth the effort.

audbackend/core/artifactory.py
frankenjoe: And there is a `explode_archive_atomic` argument to the `deploy()` function:

image

I wonder if this would ensures an artifact is only uploaded when the operation was successful (see #108). If yes and if it does not have any negative effects, we should probably turn it on. This would avoid a user has to ask an administrator to delete corrupted files if she's lacking permission to delete artifacts.

audbackend/core/artifactory.py
```diff
@@ -1,13 +1,143 @@
...
+
+        config_entry = artifactory.get_global_config_entry(host)
```
frankenjoe, line 55: I checked and the function already takes care of removing `http://` or `https://`.

audbackend/core/artifactory.py
frankenjoe: Actually, there is `artifactory.get_config_entry()` function that allows a user to set the config path. I'll set WIP and see if we can use it instead of `artifactory.get_global_config_entry()` to control the location of the config file.

audbackend/core/artifactory.py
hagenw: But doesn't `explode_archive_atomic` only apply when `explode_archive` is `True`?
And if I understand it correctly storing an archive exploded means it is stored as a folder containing all the files and not as an archive, compare https://stackoverflow.com/a/6980228.

audbackend/core/artifactory.py
frankenjoe: Done, the path of the config file can now be controlled with `ARTIFACTORY_CONFIG_FILE`. Will remove the WIP.

audbackend/core/artifactory.py
frankenjoe: Ok, too bad they don't support deploying a regular file as an atomic operation.

hagenw: Regarding the checksums, you are most likely right. When I interpret https://jfrog.com/help/r/jfrog-rest-apis/deploy-artifact-by-checksum correctly there is no need at all to provide any checksums when deploying. So I guess it will be fine to just pass on the MD5 sum.

audbackend/core/artifactory.py
```diff
@@ -1,30 +1,158 @@
...
+    if api_key is None or username is None:
+
+        config = artifactory.read_config(config_file)
+        config_entry = artifactory.get_config_entry(config, host)
+
+        if config_entry is not None:
+            if username is None:
+                username = config_entry['username']
+            if api_key is None:
+                api_key = config_entry['password']
```
hagenw, line 45: Those entries might not be part of the config given for that host. It will be safer to change to:

```suggestion
            if username is None:
                username = config_entry.get('username', None)
            if api_key is None:
                api_key = config_entry.get('password', None)
```

frankenjoe: Good point!

frankenjoe: I also added a test for it.

tests/test_artifactory.py
```diff
@@ -1,15 +1,83 @@
...
+    assert backend._api_key == ''
+
+    # config file entry with username and password
+
+    username = 'bad'
+    api_key = 'bad'
+    with open(config_path, 'w') as fp:
+        fp.write(f'[{host}]\n')
+        fp.write(f'username = {username}]\n')
+        fp.write(f'password = {api_key}]\n')
```
hagenw, line 66:
```suggestion
        fp.write(f'username = {username}\n')
        fp.write(f'password = {api_key}\n')
```

PR #112, opened 2023-04-27
get_file(): skip if file with same checksum exists

### LINKED ISSUE #109: Get file from backend only if checksums differ
In `Backend.put_file()` we already check if the checksum of a local and backend file is the same and in that case we skip the operation. We can do the same with `Backend.get_file()`.

frankenjoe:
Closes #109 

image

### DIFF
--- a/audbackend/core/backend.py
+++ b/audbackend/core/backend.py
@@ -231,6 +231,9 @@ class Backend:
     ) -> str:
         r"""Get file from backend.
 
+        The operation is silently skipped,
+        if a local file with the same checksum already exists.
+
         Args:
             src_path: path to file on backend
             dst_path: destination path to local file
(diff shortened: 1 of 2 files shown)


audbackend/core/backend.py
```diff
@@ -263,19 +266,23 @@ def get_file(
 
         audeer.mkdir(dst_root)
         if (
-                not os.access(dst_root, os.W_OK) or
-                (os.path.exists(dst_path) and not os.access(dst_root, os.W_OK))
+            not os.access(dst_root, os.W_OK) or
+            (os.path.exists(dst_path) and not os.access(dst_root, os.W_OK))
```
hagenw, line 270: Why do we need the second line? If this is true, the first line should always be true as well.

hagenw (commented):
What did we before, skip if file exists or overwrite it?

frankenjoe:
> What did we before, skip if file exists or overwrite it?

We always overwrite the file even if the files were identical. Now we only overwrite if the checksums do not match.

audbackend/core/backend.py
frankenjoe: Thanks for spotting, the second line should check if we have permission to `dst_path`, so I fixed it to:

```
(os.path.exists(dst_path) and not os.access(dst_path, os.W_OK))
```

I also added a test for it now.

PR #115, opened 2023-05-02
Calculate md5 with audeer

frankenjoe:
This removes `audbackend.md5`. 

Usually we would probably deprecate it first, but since we have so many breaking changes, I guess it's fine to remove it.


hagenw:
Yes, we will anyway have breaking changes, so I would just focus on designing `audbackend` the best way possible and remove any clutter left.

PR #116, opened 2023-05-02
DOC: remove authentication section

### LINKED ISSUE #114: DOC: update authentication section
In the authentication section we still link to `audfactory`, but since we don't rely on the package any longer we should probably directly include this information in the documentation `audbackend`.

image

frankenjoe:
Closes #114

### DIFF
--- a/docs/authentication.rst
+++ b/docs/authentication.rst
@@ -1,9 +1,0 @@
-Authentication
-==============
-
-To use the :class:`audbackend.Artifactory` backend
-you might need to authenticate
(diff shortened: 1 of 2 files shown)


hagenw:
Tests are failing as `audeer.extract_archive()` now returns the relative file paths as Windows paths. Please fix this first in a different pull request.

frankenjoe:
It is fixed in https://github.com/audeering/audbackend/pull/115

frankenjoe:
But maybe you it's indeed better to do it in a separate PR.

Issue #88, opened 2023-04-18
put_archive(): make `files` an optional argument


frankenjoe:
With `put_archive()` we currently force the user to select the files that should be included in an archive. But maybe we should make `files` an optional argument and by default include all files under `src_root` instead?

hagenw:
So basically changing
```python
Backend.put_archive(src_root, files, dst_path, version, *, tmp_root=None, verbose=False)
```
to
```python
Backend.put_archive(src_root, dst_path, version, *, files=None, tmp_root=None, verbose=False)
```
and
```
    files: if not ``None`` only the files provided here
        with their relative path from ``src_root``
        are included
```

I like the idea, as ``put_archive()`` becomes then symmetrical to ``get_archive()`` as we have with ``put_file()`` and ``get_file()``.

BTW, how to we handle relative paths that are located outside the folder at the moment?
E.g.
```python
Backend.put_archive('folder', '../a.txt', ...)
```

frankenjoe:
I wonder if make sense we first add support for `files=None` in `audeer.create_archive()`?

hagenw:
Why not, but then you should be fast: https://github.com/audeering/audeer/pull/100 :)

frankenjoe:
> BTW, how to we handle relative paths that are located outside the folder at the moment?

I checked and there are no checks currently in `create_achive()`, see https://github.com/audeering/audeer/issues/101

frankenjoe:
I also created https://github.com/audeering/audeer/issues/102 which suggests supporting `files=None`.

PR #118, opened 2023-05-02
Backend.put_archive(): make `files` an optional argument

frankenjoe:
Closes https://github.com/audeering/audbackend/issues/88

image

### DIFF
--- a/audbackend/core/backend.py
+++ b/audbackend/core/backend.py
@@ -501,7 +501,7 @@ class Backend:
         Examples:
             >>> backend.exists('/a.tar.gz', '1.0.0')
             False
-            >>> backend.put_archive('.', ['src.pth'], '/a.tar.gz', '1.0.0')
+            >>> backend.put_archive('.', '/a.tar.gz', '1.0.0')
             >>> backend.exists('/a.tar.gz', '1.0.0')
             True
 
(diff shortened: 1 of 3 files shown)


audbackend/core/backend.py
```diff
@@ -489,31 +493,23 @@ def put_archive(
...
 
-        if not os.path.exists(src_root):
-            utils.raise_file_not_found_error(src_root)
-
-        files = audeer.to_list(files)
-
-        for file in files:
-            path = os.path.join(src_root, file)
-            if not os.path.exists(path):
-                utils.raise_file_not_found_error(path)
```
hagenw, line 515: I guess you were able to remove this code as those errors are now raised by `audeer.create_archive()`?

frankenjoe: yes

tests/test_backend.py
```diff
@@ -9,70 +9,85 @@
...
-        (
+        (  # all files
             ['file.ext', 'dir/to/file.ext'],
-            'not-empty.zip',
-            'group',
-            '2.0.0',
+            '/archive.zip',
+            None,
             'tmp',
+            ['dir/to/file.ext', 'file.ext'],
```
hagenw, line 65: Could you add a test here with `files=['file.ext', 'dir/to/file.ext']`. If I remember correctly, we should then have `expected=['file.ext', 'dir/to/file.ext']`.

frankenjoe: done

hagenw: Not directly you added a test for:
```
files=['dir/to/file.ext', 'file.ext']
expected=['dir/to/file.ext', 'file.ext']
```
but I asked for
```
files=['file.ext', 'dir/to/file.ext']
expected=['file.ext', 'dir/to/file.ext']
```
to see if the order of the provided files has an influence on expected paths.
So maybe just add another test with the reversed list.

frankenjoe: Ah yes, my bad. I fixed it.

PR #119, opened 2023-05-02
DOC: document when folder is created

frankenjoe:
Adds to the docstring when a folder is created:

image

image

### DIFF
--- a/audbackend/core/backend.py
+++ b/audbackend/core/backend.py
@@ -162,6 +162,9 @@ class Backend:
         The archive type is derived from the extension of ``src_path``.
         See :func:`audeer.extract_archive` for supported extensions.
 
+        If ``dst_root`` does not exist,
+        it is created.
+
         Args:
             src_path: path to archive on backend
             dst_root: local destination directory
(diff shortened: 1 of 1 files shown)


audbackend/core/backend.py
```diff
@@ -232,8 +235,15 @@ def get_file(
...
         r"""Get file from backend.
 
-        The operation is silently skipped,
-        if a local file with the same checksum already exists.
+        If the folder of
+        ``dst_path`` does not exist,
+        it is created.
+
+        If ``dst_path`` exists
+        with a different checksum,
```
hagenw, line 243: This does mean we also replace an existing folder by a file, correct?
If yes, we should highlight this more as this might not be what a user expects.

frankenjoe: I don't think a file system allows to replace a folder with a file.

hagenw: I guess then we need to add that error to the "Raises" section.

frankenjoe: But to be on the safe side we should probably raise an error if the folder exists, which we are currently not doing.  I created https://github.com/audeering/audbackend/issues/121

PR #123, opened 2023-05-03
put_archive(): raise NotADirectoryError

frankenjoe:
Insert missing `NotADirectoryError` in raises section of `Backend.put_archive()` and adds a test for it. Also adds the same test for `Backend.get_archive()`.

image

### DIFF
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -271,12 +271,12 @@ def test_errors(tmpdir, backend):
     )
     with pytest.raises(RuntimeError, match=error_msg):
         backend.get_archive('/malformed.zip', tmpdir, version)
-    # no write permissions to `dst_path`
+    # no write permissions to `dst_root`
     if not platform.system() == 'Windows':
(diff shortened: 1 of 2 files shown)


tests/test_backend.py
```diff
@@ -339,6 +346,9 @@ def test_errors(tmpdir, backend):
             version,
             files=local_file,
         )
+    # `src_root` is not a directory
+    with pytest.raises(NotADirectoryError, match=error_not_a_folder):
+        backend.put_archive(local_path, archive, version)
```
hagenw, line 351: Was this test missing here?

frankenjoe: yes

frankenjoe:
@hagenw seems we have a bad connection to our Artifactory server today, but finally all tests were passing.

hagenw:
I created https://github.com/audeering/audbackend/pull/125 to make it easier to clean up Artifactory.

PR #126, opened 2023-05-03
Add single folder backend

### LINKED ISSUE #41: Implement backend in a single folder
This picks up a comment by @hagenw asking if every backend has to have a concept of folders: https://github.com/audeering/audbackend/issues/37#issuecomment-1114765181

Turns out, the answer is no, as the following implementation of a `SingleFolderBackend` shows (it already follows the suggestion in  #40 and implements `find()` instead of `ls()` and `glob()`).

```python
import fnmatch
import os
import shelve
import shutil
import typing

import audbackend
import audeer


class SingleFolderBackend(audbackend.Backend):

    def __init__(
            self,
            host: str,
            repository: str,
    ):
        super().__init__(audeer.safe_path(host), repository)

        self.root = audeer.mkdir(audeer.path(host, repository))
        self.map = shelve.open(audeer.safe_path(self.root, '.map'))

    def __del__(self):
        self.map.close()

    def _checksum(
            self,
            path: str,
    ) -> str:
        r"""MD5 checksum of file on backend."""
        return audbackend.core.utils.md5(path)

    def _inverse_path(
            self,
            path: str,
    ) -> (str, str, str):
        r"""Virtual file path."""
        uid = path[:8]
        folder = self.map[uid]
        version = path[9:].split('-')[-1]
        file = path[9:-len(version) - 1]
        return folder, file, version

    def _path(
            self,
            folder: str,
            name: str,
            ext: str,
            version: str,
    ) -> str:
        r"""File path on backend."""
        uid = self._uid(folder)
        self.map[uid] = folder
        file = f'{uid}-{name}{ext}-{version}'
        path = audeer.path(self.root, file)
        return path

    def _exists(
            self,
            path: str,
    ) -> bool:
        return os.path.exists(path)

    def _get_file(
            self,
            src_path: str,
            dst_path: str,
            verbose: bool,
    ):
        r"""Get file from backend."""
        shutil.copy(src_path, dst_path)

    def find(
            self,
            *,
            version: str = None,
            folder: str = None,
            pattern: str = None,
    ):
        r"""Will be implemented in BAckend."""
        folder = folder or ''
        ls = self._find(folder)
        match = set()
        for path in ls:
            f, n, v = self._inverse_path(path)
            if version is not None and v != version:
                continue
            if pattern is not None and not fnmatch.fnmatch(n, pattern):
                continue
            match.add(self.join(f, n))
        return sorted(list(match))

    def _find(
            self,
            folder: str,
    ):
        r"""List content of path."""
        ls = []
        for key, value in self.map.items():
            if value.startswith(folder):
                path = audeer.path(self.root, f'{key}*')
                match = audeer.list_file_names(path, basenames=True)
                ls.extend(match)
        return ls

    def _put_file(
            self,
            src_path: str,
            dst_path: str,
            verbose: bool,
    ):
        r"""Put file to backend."""
        shutil.copy(src_path, dst_path)

    def _remove_file(
            self,
            path: str,
    ):
        r"""Remove file from backend."""
        os.remove(path)

    def _uid(
            self,
            folder: str,
    ) -> str:
        r"""Convert folder to unique ID."""
        return audeer.uid(from_string=folder)[:8]

    def _versions(
            self,
            folder: str,
            name: str,
    ) -> typing.List[str]:
        r"""Versions of a file."""
        uid = self._uid(folder)
        pattern = f'{uid}-{name}*'
        path = audeer.path(self.root, pattern)
        vs = []
        for file in audeer.list_file_names(path):
            v = file.split('-')[-1]
            vs.append(v)
        return vs
```

Here's a test:

```
backend = SingleFolderBackend('.', 'sfb')

ret = backend.put_file('test.txt', 'sub/dir/test', '1.0.0')
print(ret)
ret = backend.put_file('test.txt', 'test.txt', '1.0.0')
print(ret)
ret = backend.put_file('test.txt', 'sub/dir/test', '1.1.0')
print(ret)
ret = backend.exists('sub/dir/test', '1.0.0')
print(ret)
ret = backend.exists('sub/dir/test', '2.0.0')
print(ret)
backend.get_file('sub/dir/test', 'check.txt', '1.0.0')
ret = backend.versions('sub/dir/test')
print(ret)
ret = backend.find()
print(ret)
ret = backend.find(folder='sub')
print(ret)
ret = backend.find(version='1.1.0')
print(ret)
ret = backend.find(pattern='*.txt')
print(ret)
```
```
.../sfb/3c6066d7-test-1.0.0
.../sfb/d41d8cd9-test.txt-1.0.0
.../sfb/3c6066d7-test-1.1.0
True
False
['1.0.0', '1.1.0']
['sub/dir/test', 'test.txt']
['sub/dir/test']
['sub/dir/test']
['test.txt']
```

frankenjoe:
Closes #41

Adds a backend implemented in a single folder as a proof of concept that the new interface proposed in  #46 does not require a folder structure.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,11 +7,14 @@ import audeer
 
 import audbackend
 
+from singlefolder import SingleFolder
+
 
 # list of backends that will be tested by default
 pytest.BACKENDS = [
     'artifactory',
     'file-system',
+    'single-folder',
 ]
 
 # UID for test session
@@ -20,11 +23,17 @@ pytest.BACKENDS = [
 pytest.UID = audeer.uid()[:8]
 
 
+@pytest.fixture(scope='package', autouse=True)
+def register_single_folder():
+    audbackend.register('single-folder', SingleFolder)
+
+
 @pytest.fixture(scope='package', autouse=False)
 def hosts(tmpdir_factory):
     return {
         'artifactory': 'https://audeering.jfrog.io/artifactory',
         'file-system': str(tmpdir_factory.mktemp('host')),
+        'single-folder': str(tmpdir_factory.mktemp('host')),
     }
 
 
--- a/tests/singlefolder.py
+++ b/tests/singlefolder.py
@@ -1,0 +1,170 @@
+import os
+import shelve
+import shutil
+import threading
+import typing
+
+import audeer
+
+import audbackend
+
+
+class SingleFolder(audbackend.Backend):
+    r"""Backend implemented in a single folder.
+
+    Files put on the backend
+    are stored under a random file name.
+    A serialized dictionary
+    stores the dependency between
+    backend path and the names.
+    It also stores the version
+    and checksum for every file.
+
+    """
+    class Map:
+        r"""Provides exclusive access to the map file."""
+
+        def __init__(
+                self,
+                path: str,
+                lock: threading.Lock,
+                *,
+                flag: str = 'w',
+        ):
+            self.obj = shelve.open(
+                path,
+                flag=flag,
+                writeback=True,
+            )
+            self.lock = lock
+
+        def __enter__(self):
+            self.lock.acquire()
+            return self.obj
+
+        def __exit__(
+                self,
+                type,
+                value,
+                traceback,
+        ):
+            self.obj.close()
+            self.lock.release()
+
+    def __init__(
+            self,
+            host: str,
+            repository: str,
+    ):
+        super().__init__(host, repository)
+
+        self._root = audeer.mkdir(audeer.path(host, repository))
+        self._path = audeer.path(self._root, '.map')
+        self._lock = threading.Lock()
+
+    def _access(
+            self,
+    ):
+        if not os.path.exists(self._path):
+            raise audbackend.core.utils.raise_file_not_found_error(self._path)
+
+        with self.Map(self._path, self._lock):
+            pass
+
+    def _checksum(
+            self,
+            path: str,
+            version: str,
+    ) -> str:
+        with self.Map(self._path, self._lock) as m:
+            return m[path][version][1]
+
+    def _create(
+            self,
+    ):
+        if os.path.exists(self._path):
+            raise audbackend.core.utils.raise_file_exists_error(self._path)
+        with self.Map(self._path, self._lock, flag='n'):
+            pass
+
+    def _delete(
+            self,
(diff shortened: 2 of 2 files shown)


frankenjoe:
@hagenw I guess it's enough we add it to the tests or you think we should add it to the package but maybe not register by default?

hagenw:
I would only add it to the tests.

tests/conftest.py
```diff
@@ -20,12 +23,21 @@
 pytest.UID = audeer.uid()[:8]
 
 
+@pytest.fixture(scope='package', autouse=True)
+def register_single_folder():
+    if os.name != 'nt':
+        audbackend.register('single-folder', SingleFolder)
```
hagenw, line 29: Could you add a comment why it fails under Windows

frankenjoe: Windows is already giving me too much pain today.

tests/singlefolder.py
```diff
@@ -0,0 +1,170 @@
...
+class SingleFolder(audbackend.Backend):
+    r"""Backend implemented in a single folder.
+
+    Files put on the backend
+    are stored under a random file name.
+    A serialized dictionary
+    stores the dependency between
+    backend path and the names.
+    It also stores the version
+    and checksum for every file.
```
hagenw, line 21: Just to see if I understood it correctly, if we put `file.txt` with version `1.0.0`, modify it and put it with version `2.0.0` we will have a mapping like

| backend path | version | checksum | random file name |
| --- | --- | --- | --- |
| `file.txt` | `1.0.0` | `<checksum0>` | `adnjedilhjgrpgjop` |
| `file.txt` | `2.0.0` | `<checksum1>` | `asjidofkwkeptokmd` |

whereas the random file name is used to store the file on the backend.

frankenjoe: yes

tests/conftest.py
hagenw: Hehe, ok, don't worry. I just thought you know as always ;)

tests/singlefolder.py
hagenw: Cool, makes sense to me and should be sufficient as a proof that we support single folder backends.

frankenjoe: I changed the title of the PR accordingly.

PR #128, opened 2023-05-04
Add Backend.date()

### LINKED ISSUE #75: Add author() and date() methods
It is interesting to see who and when a file was uploaded or changed on the server.
At the moment we usually achieve this by using custom code for Artyifactory, e.g.

```python
url = f'{host}/{repository}/{name}/db/{version}/db-{version}.zip'
path = audfactory.path(url)
stat = path.stat()
published_by = stat.created_by
published_at = f'{stat.ctime:%Y-%m-%d}'
```

But it would be nice if we could abstract this code from the backend.

frankenjoe:
Closes #75 

image

### DIFF
--- a/audbackend/core/backend.py
+++ b/audbackend/core/backend.py
@@ -96,6 +96,58 @@ class Backend:
         """
         raise NotImplementedError()
 
+    def _date(
+            self,
+            path: str,
+            version: str,
+    ) -> str:  # pragma: no cover
+        r"""Get date of file on backend.
+
+        * Return empty string if date cannot be determined
+        * Format should be '%Y/%m/%d'
+
+        """
+        raise NotImplementedError()
+
+    def date(
+            self,
+            path: str,
+            version: str,
+    ) -> str:
+        r"""Get last modification date of file on backend.
+
+        If the date cannot be determined,
+        an empty string is returned.
+
+        Args:
+            path: path to file on backend
+            version: version string
+
+        Returns:
+            date in format ``'YY/mm/dd'``
+
+        Raises:
+            BackendError: if an error is raised on the backend,
+                e.g. ``path`` does not exist
+            ValueError: if ``path`` does not start with ``'/'`` or
+                does not match ``'[A-Za-z0-9/._-]+'``
+            ValueError: if ``version`` is empty or
+                does not match ``'[A-Za-z0-9._-]+'``
+
+        Examples:
+              >>> backend.date('/f.ext', '1.0.0')
+              '1991/02/20'
+
+        """
+        path = utils.check_path(path)
+        version = utils.check_version(version)
+
+        return utils.call_function_on_backend(
+            self._date,
+            path,
+            version,
+        )
+
     def _delete(
             self,
     ):  # pragma: no cover
--- a/audbackend/core/utils.py
+++ b/audbackend/core/utils.py
@@ -1,3 +1,4 @@
+import datetime
 import errno
 import os
 import re
@@ -70,6 +71,10 @@ def check_version(version: str) -> str:
         )
 
     return version
+
+
+def date_format(date: datetime.datetime) -> str:
+    return date.strftime('%Y/%m/%d')
 
 
 def file_owner(path: str) -> str:
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -1,3 +1,4 @@
+import datetime
 import os
 import platform
 import re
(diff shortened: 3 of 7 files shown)


tests/test_backend.py
```diff
@@ -512,6 +513,8 @@ def test_file(tmpdir, src_path, dst_path, version, backend, owner):
     assert os.path.exists(src_path)
     assert backend.checksum(dst_path, version) == audeer.md5(src_path)
     assert backend.owner(dst_path, version) == owner
+    date = datetime.datetime.today().strftime('%Y/%m/%d')
+    assert backend.date(dst_path, version) == date
```
frankenjoe, line 517: It means tests can fail if executed just before midnight, but this should not happen very often :)

audbackend/core/backend.py
```diff
@@ -96,6 +96,58 @@ def _create(
...
 
+    def _date(
+            self,
+            path: str,
+            version: str,
+    ) -> str:  # pragma: no cover
+        r"""Get date of file on backend.
+
+        * Return empty string if date cannot be determined
+        * Format should be '%Y/%m/%d'
```
hagenw, line 107: Should we move `utils.date_format(date)` and request date is returned as `datetime` object or is this more ambiguous?

tests/test_backend.py
hagenw: Should indeed be fine

audbackend/core/backend.py
frankenjoe: I though it's easier this way as otherwise the function has to return `None` if it cannot get the date.

audbackend/core/utils.py
```diff
@@ -72,6 +73,10 @@ def check_version(version: str) -> str:
     return version
 
 
+def date_format(date: datetime.datetime) -> str:
+    return date.strftime('%Y/%m/%d')
```
frankenjoe, line 77: I checked in `audmodel` we actually return with `-` instead of `/`:

image

audbackend/core/backend.py
frankenjoe: So I would rather stick with the current implementation or otherwise return the `datetime.datetime` object also with `Backend.date(). But then we have to support `None` as return value.

audbackend/core/backend.py
```diff
@@ -96,6 +96,58 @@ def _create(
...
+            BackendError: if an error is raised on the backend,
+                e.g. ``path`` does not exist
+            ValueError: if ``path`` does not start with ``'/'`` or
+                does not match ``'[A-Za-z0-9/._-]+'``
+            ValueError: if ``version`` is empty or
+                does not match ``'[A-Za-z0-9._-]+'``
+
+        Examples:
+              >>> backend.date('/f.ext', '1.0.0')
+              '1991/02/20'
```
frankenjoe, line 139: Oh, I just realize that this is not ``'YY/mm/dd'``, but maybe ```'YYYY/mm/dd'`` indeed makes more sense.

audbackend/core/backend.py
hagenw: OK, no let's stick with string then. It feels also better to me that `_date()` has the same return type as `date()`

audbackend/core/backend.py
hagenw: Yes, let's go with `'YYYY/mm/dd'`. BTW, has the uppercase letter a meaning or could we also write `'yyyy/mm/dd'`?

audbackend/core/utils.py
hagenw: Using `-` or `/` seems just a convention to me. Do you have a preference?

audbackend/core/backend.py
frankenjoe: I changed to `'yyyy/mm/dd'`.

audbackend/core/utils.py
frankenjoe: According to https://docs.python.org/3/library/datetime.html#datetime.date.strftime `-` seems to be the isoformat:

image

So maybe change to `-` then?

hagenw: There seems to be an ISO standard using `-`: https://en.wikipedia.org/wiki/ISO_8601

frankenjoe: done

audbackend/core/backend.py
hagenw, line 107:
```suggestion
        * Format should be '%Y-%m-%d'
```

Issue #129, opened 2023-05-09
Provide script to convert repository structures


hagenw:
In `audbackend` we will introduce a new structure in which files are stored on the backend (as added in https://github.com/audeering/audbackend/pull/62), changing from

```
/<folder>/<name-wo-ext>/<version>/<name-wo-ext>-<version>.<ext>
```
to
```
/<sub-path>/<version>/<name>
```

As discussed in #76 it would be useful to provide a script that can convert an existing repository structure to a new one, for example by copying the files from one repository to another one.

When dealing with Artifactory as a backend it might be more safe to first download all files, convert them, and upload them afterwards.

frankenjoe:
With https://github.com/audeering/audbackend/pull/130 there is no longer a need to convert the repository file structure. We now offer a solution to keep working with repositories that use the old structure.

hagenw:
I agree, there is no need for a script and we should focus on #131 instead.

PR #130, opened 2023-05-12
Artifactory: support old file structure

### LINKED ISSUE #76: How to migrate/include legacy repositories
As discussed in https://github.com/audeering/audbackend/issues/58 the only solution to removing the `ext` argument from all our method requires that we store artifacts differently on the servers (see https://github.com/audeering/audbackend/pull/62), namely:

```
<host>/<repository>/<folder>/<version>/<name>
```

instead of

```
<host>/<repository>/<folder>/<name-wo-ext>/<version>/<name-wo-ext>-<version>.<ext>
```

This means we need a strategy how to handle already existing repositories that store artifacts in the old structure.
One solution was already proposed in https://github.com/audeering/audbackend/issues/58 which we follow up on here.

Proposal:

* Rename existing repositories and add `-legacy` at their end, e.g. `data-public-local-legacy` and make them read only (might need to be temporarily changed if data has to be deleted)
* Make sure the affected Python packages are able to read from different repositories
* Implement an `artifactory-legacy` backend in `audbackend`
* Register the old and new repositories with the Python packages, e.g. with an entry in `.audb.yaml`
* Add an [upload exclude pattern](https://jfrog.com/knowledge-base/how-to-use-include-exclude-patterns/) to the new repositories on Artifactory to make sure users that use an old version of `audbackend` are not able to upload new artifacts.
* Add an [upload exclude pattern](https://jfrog.com/knowledge-base/how-to-use-include-exclude-patterns/) to the legacy repositories on Artifactory to make sure users can only upload using the `artifactory-legacy` backend or an old version of `audbackend` (they repos should be readonly, but it might be needed to access files to delete single artifacts).

frankenjoe:
Closes #76 #129 

With this PR we add support for storing files on Artifactory using the old structure, i.e.:

`<host>/<repository>/<folder>/<name-wo-ext>/<version>/<name-wo-ext>-<version>.<ext>`

In packages where we want to stick with the old structure we can then do:

```python
backend = audbackend.access(...)
if isinstance(backend, audbackend.Artifactory):
    backend._use_legacy_file_structure(extensions=...)
```

Other changes:

* Checks if config file exists before trying to read credentials from it

### DIFF
--- a/audbackend/core/artifactory.py
+++ b/audbackend/core/artifactory.py
@@ -304,7 +304,12 @@ class Artifactory(Backend):
             )
             if not path.exists():
                 utils.raise_file_not_found_error(str(path))
-            paths = [str(x) for x in path.glob("**/*") if x.is_file()]
+            path = str(path)
+
+            # listing all path on repository
+            # is faster than path.glob('**/*')
+            paths = [str(x) for x in self._repo]
+            paths = [x for x in paths if x.startswith(path)]
 
         else:  # find versions of path
 
--- a/audbackend/core/artifactory.py
+++ b/audbackend/core/artifactory.py
@@ -159,7 +159,7 @@ class Artifactory(Backend):
 
         # to support legacy file structure
         # see _use_legacy_file_structure()
-        self._legacy_extensions = None
+        self._legacy_extensions = []
         self._legacy_file_structure = False
 
     def _access(
--- a/audbackend/core/artifactory.py
+++ b/audbackend/core/artifactory.py
@@ -154,37 +154,10 @@ class Artifactory(Backend):
         )
         self._repo = path.find_repository_local(self.repository)
 
-        self.extensions = []
-        r"""Custom extensions.
-
-        On Artifactory files are stored
-        under the following structure:
-        ``'.../<name-wo-ext>/<version>/<name-wo-ext>-<version>.<ext>'``.
-        By default,
-        the extension
-        ``<ext>``
-        is set the string after the last dot.
-        I.e.,
-        the backend path
-        ``'.../file.tar.gz'``
-        will translate into
-        ``'.../file.tar/1.0.0/file.tar-1.0.0.gz'``.
-        However,
-        by defining custom extensions
-        it is possible to overwrite
-        the default behavior.
-        E.g.,
-        with
-        ``backend.extensions.append('tar.gz')``
-        it is ensured that
-        ``'tar.gz'``
-        will be recognized as an extension
-        and the backend path
-        ``'.../file.tar.gz'``
-        will then translate into
-        ``'.../file/1.0.0/file-1.0.0.tar.gz'``.
-
-        """
+        # to support legacy file structure
+        # see _use_legacy_file_structure()
+        self._legacy_extensions = None
+        self._legacy_file_structure = False
 
     def _access(
             self,
@@ -291,6 +264,30 @@ class Artifactory(Backend):
         src_path = self._path(src_path, version)
         _download(src_path, dst_path, verbose=verbose)
 
+    def _legacy_split_ext(
+            self,
+            name: str,
+    ) -> typing.Tuple[str, str]:
+        r"""Split name into basename and extension."""
+
+        ext = None
+        for custom_ext in self._legacy_extensions:
+            # check for custom extension
+            # ensure basename is not empty
+            if name[1:].endswith(f'.{custom_ext}'):
+                ext = custom_ext
+        if ext is None:
+            # if no custom extension is found
+            # use last string after dot
+            ext = audeer.file_extension(name)
+
+        base = audeer.replace_file_extension(name, '', ext=ext)
+
+        if ext:
+            ext = f'.{ext}'
+
+        return base, ext
+
     def _ls(
             self,
             path: str,
@@ -312,8 +309,13 @@ class Artifactory(Backend):
         else:  # find versions of path
 
             root, name = self.split(path)
-            base, _ = self._split_ext(name)
-            root = f'{self._expand(root)}{base}'
+
+            if self._legacy_file_structure:
+                base, _ = self._legacy_split_ext(name)
+                root = f'{self._expand(root)}{base}'
+            else:
+                root = self._expand(root)
+
             root = _artifactory_path(
                 root,
                 self._username,
@@ -328,6 +330,12 @@ class Artifactory(Backend):
             if not paths:
                 utils.raise_file_not_found_error(path)
 
+        # <host>/<repository>/<root>/<name>
+        # ->
+        # (/<root>/<name>, <version>)
+        #
+        # or legacy:
+        #
         # <host>/<repository>/<root>/<base>/<version>/<base>-<version>.<ext>
         # ->
         # (/<root>/<base>.<ext>, <version>)
@@ -340,10 +348,15 @@ class Artifactory(Backend):
 
             name = tokens[-1]
             version = tokens[-2]
-            base = tokens[-3]
-            ext = name[len(base) + len(version) + 1:]
-            name = f'{base}{ext}'
-            path = self.sep.join(tokens[:-3])
+
+            if self._legacy_file_structure:
+                base = tokens[-3]
+                ext = name[len(base) + len(version) + 1:]
+                name = f'{base}{ext}'
+                path = self.sep.join(tokens[:-3])
+            else:
+                path = self.sep.join(tokens[:-2])
+
             path = self.sep + path
             path = self.join(path, name)
 
@@ -368,16 +381,26 @@ class Artifactory(Backend):
     ) -> artifactory.ArtifactoryPath:
         r"""Convert to backend path.
 
+        <root>/<name>
+        ->
+         <host>/<repository>/<root>/<version>/<name>
+
+        or legacy:
+
         <root>/<base>.<ext>
         ->
         <host>/<repository>/<root>/<base>/<version>/<name>-<version>.<ext>
 
         """
         root, name = self.split(path)
-        base, ext = self._split_ext(name)
-
         root = self._expand(root)
-        path = f'{root}{base}/{version}/{base}-{version}{ext}'
+
+        if self._legacy_file_structure:
+            base, ext = self._legacy_split_ext(name)
+            path = f'{root}{base}/{version}/{base}-{version}{ext}'
+        else:
+            path = f'{root}{version}/{name}'
+
         path = _artifactory_path(
             path,
             self._username,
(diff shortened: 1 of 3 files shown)


audbackend/core/artifactory.py
```diff
@@ -292,12 +305,9 @@ def _ls(
             paths = [str(self._path(path, v))
                      for v in vs if self._exists(path, v)]
 
-            if not paths:
-                utils.raise_file_not_found_error(path)
```
frankenjoe, line 296: No longer needed, since a `FileNotFoundError` will be raised in line 305 in that case.

audbackend/core/artifactory.py
```diff
@@ -154,6 +154,38 @@ def __init__(
...
+        ``backend.extensions.append('tar.gz')``
+        it is ensured that
+        ``'tar.gz'``
+        will be recognized as an extension
+        and the backend path
+        ``'.../file.tar.gz'``
+        will then translate into
+        ``'.../file/1.0.0/file-1.0.0.tar.gz'``.
+
+        """
```
hagenw, line 187: From the docstring one cannot easily infer the default value of `extensions` and that the user is supposed to provide a list of additional extensions. I would simply add 1-2 sentences stating this.

audbackend/core/artifactory.py
```diff
@@ -154,6 +154,38 @@ def __init__(
         )
         self._repo = path.find_repository_local(self.repository)
 
+        self.extensions = []
```
frankenjoe, line 157: If we think this might be needed by other backend implementations, too. We could also move it to `Backend` and leave it to the backend implementation if they use the information or not.

hagenw (commented):
This is indeed more elegant as it automatically ensures that all methods work for a custom extension, and we don't need to check for which methods we have to add legacy versions (e.g. `put_file_legacy()`).

What we did not cover in our discussion in https://github.com/audeering/audbackend/issues/76 is how to handle the different behavior compared to other backends. If a developer uses `Artifactory.extensions` in their code then they will no longer be able to switch to another backend.
One solution would be to not add `Artifactory.extensions` to the official API, and only provide it for us to use inside `audmodel` and `audbenchmark`.

The other solution would be to add it to every backend. This would then allow us to also run unit tests using the `file-system` backend in `audmodel` and `audbenchmark`.

Any preference on this?

hagenw:
> This is indeed more elegant as it automatically ensures that all methods work for a custom extension, and we don't need to check for which methods we have to add legacy versions (e.g. `put_file_legacy()`).
> 
> What we did not cover in our discussion in #76 is how to handle the different behavior compared to other backends. If a developer uses `Artifactory.extensions` in their code then they will no longer be able to switch to another backend. One solution would be to not add `Artifactory.extensions` to the official API, and only provide it for us to use inside `audmodel` and `audbenchmark`.
> 
> The other solution would be to add it to every backend. This would then allow us to also run unit tests using the `file-system` backend in `audmodel` and `audbenchmark`.
> 
> Any preference on this?

I think I'm wrong here and it would not break switching to another backend. Setting the extension for other backends would only be needed to match already existing file paths on the backend.

audbackend/core/artifactory.py
frankenjoe: The default value is an empty list and in the docstring we show an example how to add an extension to it, so I guess it should be fine.

audbackend/core/artifactory.py
frankenjoe: The default value is an empty list and in the docstring we show an example how to add an extension to it, so I guess it should be fine.

audbackend/core/artifactory.py
frankenjoe: If we decide to make it available to all backends we could make `Backend.extensions` a property that is set with `create()` and `access()`. The nicest solution would be to have it only with `create()`, but backends that use `extensions` must store them somehow with the repository (e.g. in a hidden file in the root directory).

hagenw: If I understand it correctly (compare https://github.com/audeering/audbackend/pull/130#issuecomment-1547269016), we need `Artifactort.extensions` only to handle already existing repositories that have used extensions including a dot.

In order to allow backward compatibility for all backends it would indeed make sense to provide it not only for `Artifactory`, but for `Backend`. On the other hand, this only makes sense when also switching back to the old file structure on the `file-system` backend.

As a feature for new repositories I found `extensions` a little bit risky as now a user will not only have to know the backend and repository name, but also the `Backend.extensions` settings needed to work with an existing repository. So it looks like this property would then need to be stored in the repository itself. But I would not go that far. Instead I would vote to rename it to `Backend.legacy_extensions()` and not include it in the default API documentation.
Instead I would add a section to the documentation discussing how to handle already existing repositories with the new `audbackend` version. This documentation, I would propose to add in another pull request.

hagenw: >> On the other hand, this only makes sense when also switching back to the old file structure on the file-system backend.
>
> I think it's ok if we are not backward compatible with file-system. So I would not switch back to the old structure.

I agree, and created https://github.com/audeering/audbackend/issues/131 to track documenting this.

> Ok, but then I propose we also use `<host>/<repository>/<folder>/<name-wo-ext>/<version>/<name-wo-ext>-<version>.<ext>` only on legacy repositories and otherwise switch to `<host>/<repository>/<folder>/<version>/<name>.<ext>`. Without an official way to specify custom extensions the old structure makes no sense to me since e.g. file.tar.gz will be stored as `'.../file.tar/1.0.0/file.tar-1.0.0.gz'`.

I disagree here as it again would make things more complicated then is needed.
I would always stick to `<host>/<repository>/<folder>/<name-wo-ext>/<version>/<name-wo-ext>-<version>.<ext>`,
and for the new Artifactory backend I would not provide a way to configure what is a valid file extension. This way the resulting path on the backend will always be easily predictable and not depend on a user setting.
I don't see a big issue in having `'.../file.tar/1.0.0/file.tar-1.0.0.gz'` as path.

hagenw:
This pull request currently handles two different things at once:
* switching back to the old file structure on Artifactory
* adding an attribute to configure setting valid file extensions

I would move the second part to another pull request and focus only on the first part here.

audbackend/core/artifactory.py
frankenjoe: > I don't see a big issue in having '.../file.tar/1.0.0/file.tar-1.0.0.gz' as path.

I think it can be an issue. E.g. if they provide a way to preview TAR.GZ files on Artifactory as they do with ZIP in their UI. And also generally, it doesn't feel right to use `gz` as file extension when the extension is `tar.gz`.

frankenjoe: And also don't forget that one reason we added the `ext` argument was to group files in the same directory on Artifactory, e.g. the YAML files that store header and the meta information in `audmodel`.

hagenw: The only save usage of letting the user configure possible additional extension names seems to me during the creation of the repository. After the first files have been uploaded the user shouldn't be allowed to change the file extension settings as this might lead to the case that certain files can no longer be accessed.

frankenjoe: > The only save usage of letting the user configure possible additional extension names seems to me during the creation of the repository.

Yes, and that's a strong argument we should not generally switch back to the old file structure as it only makes things complicated. Instead we keep the new file structure and only introduce a legacy function that allows to switch to the old file structure in packages where relied on it on the past.

hagenw: I guess we are stuck here again ;)

My requirement would still be that all old and new versions of `audb` should work with already existing repos without that the user needs to do anything. And so far I did not see how this will work when we switch to another file structure on the backend.

frankenjoe: But with `audb` it will use the old structure, even when we create new repositories since we hard code this on package level. Only new packages, where we don't force the legacy package will use the structure. This makes since we don't want the user to bother file extensions.

hagenw: Ah, ok, now I get it. So basically with `Artifactory._use_legacy_file_structure(extensions=[])` we can solve all the issues and it also makes sense to implement both inside this pull request.

We could add the switch-function also to the `file-system` backend then we don't need to add extra explanation how to convert existing `file-system` backends and could also continue in `audb` using the old structure.

frankenjoe: I added:

image

and updated the description accordingly.

audbackend/core/artifactory.py
frankenjoe: No longer relevant since we removed `Artifactory.extensions`.

frankenjoe:
All tests are passing again, so it seems with https://github.com/audeering/audb/pull/306 works as expected we finally found a solution that does not require any change to our existing repositories and still allows us to switch to the new backend interface :)

audbackend/core/artifactory.py
```diff
@@ -259,6 +267,30 @@ def _get_file(
...
         _download(src_path, dst_path, verbose=verbose)
 
+    def _legacy_split_ext(
+            self,
+            name: str,
+    ) -> typing.Tuple[str, str]:
+        r"""Split name into basename and extension."""
+
+        ext = None
+        for custom_ext in self._legacy_extensions:
```
hagenw, line 277: Above you set the default value to
```python
self._legacy_extensions = None
```
so it seems a little bit risky to just loop it here. We should switch to `[]` as default or add a check here if `self._legacy_extensions` is `None`

audbackend/core/artifactory.py
```diff
@@ -367,3 +432,43 @@ def _remove_file(
...
+        r"""Use legacy file structure.
+
+        Stores files under
+        ``'.../<name-wo-ext>/<version>/<name-wo-ext>-<version>.<ext>'``
+        instead of
+        ``'.../<version>/<name>'``.
+        By default,
+        the extension
+        ``<ext>``
+        is set the string after the last dot.
```
hagenw, line 450:
```suggestion
        is set to the string after the last dot.
```

audbackend/core/artifactory.py
hagenw: Now I found that `_use_legacy_file_structure()` will ensure that it is never set to `None` when the legacy paths are used. But maybe it will still be easier to understand of the default is `[]` or to we gain some additional information by using `None` instead?

tests/test_artifactory.py
```diff
@@ -125,3 +127,34 @@ def test_errors(tmpdir, backend):
...
+
+@pytest.mark.parametrize(
+    'backend',
+    ['artifactory'],
+    indirect=True,
+)
+@pytest.mark.parametrize(
+    'file, version, extensions, expected',
+    [
+        ('/file.tar.gz', '1.0.0', None, 'file.tar/1.0.0/file.tar-1.0.0.gz'),
```
hagenw, line 140: I would add one additional test here to make sure we also cover that case:

```suggestion
        ('/file.tar.gz', '1.0.0', None, 'file.tar/1.0.0/file.tar-1.0.0.gz'),
        ('/file.tar.gz', '1.0.0', [], 'file.tar/1.0.0/file.tar-1.0.0.gz'),
```

audbackend/core/artifactory.py
frankenjoe: I changed to `[]` as default value.

audbackend/core/artifactory.py
hagenw: In https://github.com/audeering/audbackend/pull/130#discussion_r1193651913 I suggested to add the legacy function also to the `FileSystem` backend as the user could then mimic how the files would be stored on the Artifactory server. But as we want to support several backends in the long run and each might store the files differently, there might not be a need to add the legacy method to `FileSystem` as well (I would anyway not do this inside this pull request).

On the other hand, if a University have used `audb` together with a NAS and the `FileSystem` backend it would still ensure backward compatibility in that case for `audb`.

What would you prefer?

frankenjoe: I don't see an advantage if we mimic the file system on the `Artifactory` with `FileSystem`. 

> On the other hand, if a University has used audb together with a NAS and the FileSystem backend it would still ensure backward compatibility in that case for audb.

What are the chances? :) But if we want to backward compatible, we could apply the same solution to `FileSystem`.

hagenw: I'm also not completely decided, but I created https://github.com/audeering/audbackend/issues/136 to track this.

PR #134, opened 2023-05-16
DOC: add usage section

### LINKED ISSUE #60: Add developer section to documentation
We should add a developer section to the documentation that explains how to implement a new backend.

frankenjoe:
Closes #60

Adds a usage section to the documentation discussing:

* How to use existing backends
* How to develop new backends

image

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -49,8 +49,8 @@ a SQLite_ database.
         return decorator
 
 
-Basic usage
------------
+Basics
+------
 
 The heart of
 :mod:`audbackend`
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -3,20 +3,23 @@ Usage
 
 The aim of
 :mod:`audbackend`
-is to provide an abstract layer
-to interface with different
-file storage systems.
-Instead of directly communicating
-with a specific file system,
-the abstract class
-:class:`audbackend.Backend`.
-This allows users to
-execute the same code
-on different systems.
-For instance,
-in this tutorial
-we create a backend that
-stores files into
+is to provide an
+abstract interface for
+any kind of file storage system.
+Even those,
+that have not been
+invented yet :)
+
+This tutorial is divided
+into two parts.
+In the first part,
+we show the basic usage
+by means of a
+standard file system.
+In the second part,
+we take a deep dive
+and develop a backend
+that stores files into
 a SQLite_ database.
 
 .. set temporal working directory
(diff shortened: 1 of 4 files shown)


docs/usage.rst
```diff
@@ -0,0 +1,719 @@
...
+is to provide an
+abstract interface for
+any kind of file storage system.
+Even those,
+that have not been
+invented yet :)
+
+This tutorial is divided
+into two parts.
+In the first part,
```
hagenw, line 15: I would directly reference the parts here.

Therefore you would need to add

```rst
.. _basics:
```
before the actual sub-section and then change the code here to:

```suggestion
Under :ref:`basics`,
```

docs/usage.rst
```diff
@@ -0,0 +1,719 @@
...
+we show the basic usage
+by means of a
+standard file system.
+In the second part,
```
hagenw, line 19: As for basics, I would directly link to the section by adding

```rst
.. _development:
```
in front of the sub-section and changing here to

```suggestion
Under :ref:`development`,
```

docs/usage.rst
```diff
@@ -0,0 +1,719 @@
...
+  (it is possible to have several repositories
+  on the same host).
+
+.. jupyter-execute::
+
+    backend = audbackend.create('file-system', './host', 'repo')
+
+
+This will create an empty repository
+(in our case the folder ``'./<host>/<repo>/'``).
```
hagenw, line 99:
```suggestion
(in our case the folder ``'./host/repo/'``).
```

docs/usage.rst
```diff
@@ -0,0 +1,719 @@
...
+    backend = audbackend.access('file-system', './host', 'repo')
+
+
+To put a file on the backend,
+we provide two path arguments.
+
+* ``src_path``: path to a file on the local file system.
+  This is the file we want to store on the backend.
+* ``dst_path``: virtual path that represents the file on the backend.
+  It serves as an alias that is understood by all backends.
```
hagenw, line 114: At some point we need to add that the path should always start with `/`. Maybe here is the right location?

docs/usage.rst
```diff
@@ -0,0 +1,719 @@
...
+as an archive.
+Here,
+we compress the file
+we just modified
+and put it under
+the sub-path ``'/a/'``.
+
+.. jupyter-execute::
+
+    backend.put_archive('.', '/a/file.zip', '2.0.0', files=[path])
```
hagenw, line 194: Why start with version 2.0.0 here?

docs/usage.rst
```diff
@@ -0,0 +1,719 @@
...
+
+
+When we get an archive from the backend
+we can automatically extract it.
+
+.. jupyter-execute::
+
+    paths = backend.get_archive('/a/file.zip', '.', '2.0.0')
```
hagenw, line 202: I guess we need to discuss here if this overwrites the already existing file or not.

hagenw (commented):
Great, thanks for writing the extensive docs.

I will start with a review on the "Basics" section.

docs/usage.rst
```diff
@@ -0,0 +1,719 @@
...
+.. jupyter-execute::
+
+    @add_method(SQLite)
+    def _get_file(
+                self,
+                src_path: str,
+                dst_path: str,
+                version: str,
+                verbose: bool,
+        ):
```
hagenw, line 571:
```suggestion
            self,
            src_path: str,
            dst_path: str,
            version: str,
            verbose: bool,
    ):
```

docs/usage.rst
```diff
@@ -0,0 +1,719 @@
...
+                '''
+                ls = db.execute(query).fetchall()
+
+        if not ls and not path == '/':
+            raise FileNotFoundError(
+                errno.ENOENT,
+                os.strerror(errno.ENOENT),
+                path,
+            )
+        return ls
```
hagenw, line 630: The two different cases and the check before the error message are not obvious.
I would suggest to add comments to the code what is going on, or discuss it in the text.

docs/usage.rst
```diff
@@ -0,0 +1,719 @@
...
+    ):
+        with self._db as db:
+            query = f'''
+                DELETE
+                FROM data
+                WHERE path="{path}" AND version="{version}"
+            '''
+            db.execute(query)
+
+    backend.remove_file('/file.txt', '2.0.0')
```
hagenw, line 664: I guess the goal here is to really remove an existing path and not do demonstrate that it will not raise an error when the file does not exist. 
```suggestion
    backend.remove_file('/file.txt', '1.0.0')
```

hagenw (commented):
And here are my remarks for the "Development" section.

hagenw:
The current naming of the subsections is a little bit to general:

image

as "Development" could also refer to the development of `audbackend`.
I would suggest to change them to:

* File-system example
* Develop new backend

frankenjoe:
@hagenw please feel free to overtake the PR.

docs/usage.rst
hagenw: Added comments to the code.

docs/usage.rst
hagenw: I switched to version 1.0.0

docs/usage.rst
hagenw: I switched to version 1.0.0

PR #144, opened 2023-09-18
Add support for Python 3.11

hagenw:
With the release of `dohq-artifactory >=0.9.1` all the remaining issues with Python 3.11 should be fixed now (compare https://github.com/devopshq/artifactory/issues/404) and we should be able to support it here as well.


hagenw:
It's failing due to https://github.com/devopshq/artifactory/issues/432

PR #151, opened 2023-10-17
Add regex support to _use_legacy_file_structure()

### LINKED ISSUE #150: Support regex with _use_legacy_file_structure()
E.g. if we have extensions with a running index on repositories that use the old file system, we have to support regex with ` _use_legacy_file_structure()`.

frankenjoe:
Closes #150

### DIFF
--- a/audbackend/core/artifactory.py
+++ b/audbackend/core/artifactory.py
@@ -1,4 +1,5 @@
 import os
+import re
 import typing
 
 import artifactory
@@ -157,6 +158,7 @@ class Artifactory(Backend):
         # see _use_legacy_file_structure()
         self._legacy_extensions = []
         self._legacy_file_structure = False
+        self._legacy_file_structure_regex = False
 
     def _access(
             self,
@@ -272,7 +274,12 @@ class Artifactory(Backend):
         for custom_ext in self._legacy_extensions:
             # check for custom extension
             # ensure basename is not empty
-            if name[1:].endswith(f'.{custom_ext}'):
+            if self._legacy_file_structure_regex:
+                pattern = f'\.({custom_ext})$'
+                match = re.search(pattern, name[1:])
+                if match:
+                    ext = match.group(1)
+            elif name[1:].endswith(f'.{custom_ext}'):
                 ext = custom_ext
         if ext is None:
             # if no custom extension is found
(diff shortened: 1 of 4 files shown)


audbackend/core/artifactory.py
```diff
@@ -431,6 +438,7 @@ def _use_legacy_file_structure(
             self,
             *,
             extensions: typing.List[str] = None,
+            regex: bool = False,
```
hagenw, line 441: What happens if you mix regex based extensions with non-regex based, e.g. `extensions=['\d+.tar.gz', 'tar.gz']`?

hagenw: I'm asking as it seems a little bit critical, that we can provide a list with extensions, some using a regex pattern others not, but we can only provide a single `regex=True`.

frankenjoe: You are not supposed to mix. Either you provide all as regex or none.

tests/test_filesystem.py
```diff
@@ -61,24 +61,61 @@ def test_get_file_interrupt(tmpdir, bad_file_system, backend):
...
+            'file.0.tar/1.0.0/file.0.tar-1.0.0.gz',
+        ),
+        (
+            '/file.0.tar.gz', '1.0.0', [r'\d+.tar.gz'], True,
+            'file/1.0.0/file-1.0.0.0.tar.gz',
+        ),
+        (
+            '/file.99.tar.gz', '1.0.0', [r'\d+.tar.gz'], True,
+            'file/1.0.0/file-1.0.0.99.tar.gz',
+        ),
```
hagenw, line 110: I would add a test for our actual use case:

```suggestion
        ),
        (
            '/file.prediction.99.tar.gz', '1.0.0', [r'prediction.\d+.tar.gz', r'truth.tar.gz'], True,
            'file/1.0.0/file-1.0.0.prediction.99.tar.gz',
        ),
        (
            '/file.truth.tar.gz', '1.0.0', [r'prediction.\d+.tar.gz', r'truth.tar.gz'], True,
            'file/1.0.0/file-1.0.0.truth.tar.gz',
        ),
```

What still does not work for me is to mix TAR.GZ with 99.TAR.GZ, but maybe I just define the wrong search pattern: `r'(\d.)?tar.gz'`?

Issue #155, opened 2023-11-09
Use backend without versioning


frankenjoe:
Though the main motivation to introduce this packages was versioning. There might the use case where you don't care about versioning and want to use it like a regular file system. At the moment this is not possible since we enforce versioning. I wonder if it would make sense to support both?

hagenw:
I'm afraid this will not be so easy to achieve.

You could try to hide `version` from the user by making it a keyword argument with the default value `version='1.0.0'`. If you also want to avoid that the version is visible on the actual backend (in how the files are stored), I guess you would need additional backend classes.

frankenjoe:
Yes, at the moment what I did was creating a new project, where I removed all the version specific handling. The result was a simplified code, that still looks very similar to the original one, except that a file `/sub/file.txt` is directly stored under `/sub/file.txt` and not translated to something like `sub/1.0.0/file.txt`.

In case we wanted to support both, we could implement it by introducing a new base class that does not version and which all backends derive from. Our current base class `Backend` class would then derive from it, but before calling the functions of the base class transform `/sub/file.txt` to  `/sub/<version>/file.txt`. As a side-effect this would also simplify the implementation of a backend, since the backend implementation does not have to deal with versioning at all as this is only introduced by an intermediate layer.

hagenw:
Sounds good to me, especially:

> As a side-effect this would also simplify the implementation of a backend, since the backend implementation does not have to deal with versioning at all

frankenjoe:
As pointed out in https://github.com/audeering/audbackend/issues/166#issuecomment-1899049163, we cannot simply allow `version=None`, unless we restrict `version` to follow a certain format. Another solution is to introduce two interfaces to a backend - one with and one without versioning. I started with an implementation of this solution and here's a first usage example:

By default we use a versioned backend, which behaves exactly as our current backend:

```python
import tempfile

import audeer

import audbackend


with tempfile.TemporaryDirectory(dir='.') as tmp:

    backend = audbackend.create('file-system', tmp, 'repo')
    file = audeer.touch(tmp, '~')

    backend.put_file(file, '/file.txt', '1.0.0')
    backend.put_file(file, '/sub/file.txt', '1.0.0')
    backend.put_file(file, '/sub/file.txt', '2.0.0')

    print(backend.ls())
    print(backend.ls('/file.txt'))
    print(backend.versions('/file.txt'))
    print(backend.ls('/sub/file.txt'))
    print(backend.versions('/sub/file.txt'))
[...]

PR #165, opened 2024-01-17
Remove version from implementation

frankenjoe:
Relates to #155 

This removes the `version` argument from the backend implementation.

Instead of forwarding `path` and `version` to the implementation, `Backend` integrates the version into the path now. This simplifies the implementation of a backend, since a developer does not have to care about storing the version (especially the implementation of `_ls()` becomes way simpler). Also the support for legacy file structure is now handled directly by `Backend` (and as a side-effect supported by any backend).

The suggested changes will allow us to implement `audbackend` in a way that it can be used without versioning as requested in #155 (not implemented here).

### DIFF
--- a/audbackend/core/backend.py
+++ b/audbackend/core/backend.py
@@ -432,7 +432,6 @@ class Backend:
             '2.0.0'
 
         """
-        path = utils.check_path(path)
         vs = self.versions(path)
         return vs[-1]
 
--- a/audbackend/core/backend.py
+++ b/audbackend/core/backend.py
@@ -469,7 +469,11 @@ class Backend:
     ) -> typing.List[str]:  # pragma: no cover
         r"""List all files under sub-path.
 
-        * If path does not exist an error should be raised
+        If ``path`` is ``'/'`` and no files exist on the repository,
+        an empty list should be returned
+        Otherwise,
+        if ``path`` does not exist or no files are found under ``path``,
+        an error should be raised.
 
         """
         raise NotImplementedError()
--- a/audbackend/core/backend.py
+++ b/audbackend/core/backend.py
@@ -85,8 +85,6 @@ class Backend:
             'd41d8cd98f00b204e9800998ecf8427e'
 
         """
-        path = utils.check_path(path)
-        version = utils.check_version(version)
         path_with_version = self._path_with_version(path, version)
 
         return utils.call_function_on_backend(
@@ -146,8 +144,6 @@ class Backend:
               '1991-02-20'
 
         """
-        path = utils.check_path(path)
-        version = utils.check_version(version)
         path_with_version = self._path_with_version(path, version)
 
         return utils.call_function_on_backend(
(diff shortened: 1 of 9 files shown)


audbackend/core/backend.py
```diff
@@ -79,11 +87,11 @@ def checksum(
         """
         path = utils.check_path(path)
         version = utils.check_version(version)
+        path_with_version = self._path_with_version(path, version)
```
hagenw, line 90: I would move 
```python
path = utils.check_path(path)
version = utils.check_version(version)
```
into `self._path_with_version()` as it seems to be called always before `self._path_with_version()` is called.

This affects other methods in `Backend` as well.

audbackend/core/backend.py
```diff
@@ -592,13 +691,35 @@ def owner(
...
+            self,
+            path: str,
+            version: str,
+    ) -> str:
+        r"""Convert to versioned path.
+
+        <root>/<name>
+        ->
+        <root>/<version>/<name>
+
```
hagenw, line 711: Would be nice, if you could add a line here to the docstring also documenting how the returned path looks if `self._legacy_file_structure` is `True`.

audbackend/core/backend.py
hagenw: In addition, `utils.check_path()` can be removed in `latest_version()` as it is handled inside the call to `versions()` then.

audbackend/core/backend.py
```diff
@@ -439,14 +444,41 @@ def latest_version(
...
     def _ls(
             self,
             path: str,
-    ) -> typing.List[typing.Tuple[str, str]]:  # pragma: no cover
-        r"""List all files under (sub-)path.
+    ) -> typing.List[str]:  # pragma: no cover
+        r"""List all files under sub-path.
 
         * If path does not exist an error should be raised
-        * If path ends on `/` it is a sub-path
```
hagenw, line 449: Is this no longer true?

hagenw (commented):
Great, it's indeed very cool that you don't have to handle `version` at all when adding a new backend.

audbackend/core/backend.py
frankenjoe: Yes, makes sense. I applied both suggestions.

audbackend/core/backend.py
frankenjoe: done

audbackend/core/backend.py
frankenjoe: I removed it because we always pass a sub-path now and in the line above the docstring now says "List all files under sub-path."

frankenjoe: However, we should mention that an empty list should be returned for an empty repository. I now changed the comment to:

```python
    def _ls(
            self,
            path: str,
    ) -> typing.List[str]:  # pragma: no cover
        r"""List all files under sub-path.

        If ``path`` is ``'/'`` and no files exist on the repository,
        an empty list should be returned
        Otherwise,
        if ``path`` does not exist or no files are found under ``path``,
        an error should be raised.

        """
        raise NotImplementedError()
``

PR #171, opened 2024-01-23
DOC: split usage into usage and developer-guide

### LINKED ISSUE #169: Hide backend in API documentation
In https://github.com/audeering/audbackend/pull/167 we introduce a new abstract interface layer between backends and the user. Which means a normal user will need to look only at the interface API. For that reason we should hide parts of the backend API from the user, e.g. all the methods that are inherited from `audbackend.Backend` in `audbackend.Artifactory` and `audbackend.FileSystem` and instead add a link to the interface section.

It might still be of interest to show the API for `audbackend.Backend` as this is needed for developer who want to create a new interface.

hagenw:
Closes #169 

This splits the "Usage" documentation into "Usage" and "Developer guide".

<details><summary>Usage</summary>

image

</details>

<details><summary>Developer guide</summary>

image

</details>

It also updates the landing page, and does no longer list the class names:

image

---

When writing the usage documentation the following things seemed slightly strange to me:

* When using `audbackend.create()` or `audbackend.access()` the backend is referred to by a string name whereas the interface needs to be given as a class. Providing it as a class makes things easier, but as a user you might be wonder why there is this difference.
* `audbackend.available()` does only list backends, not interfaces
* `audbackend.available()` can only provide information on the backend class associated with a backend name, once an instance of the backend is created.

### DIFF
--- a/docs/developer-guide.rst
+++ b/docs/developer-guide.rst
@@ -324,6 +324,7 @@ stored on our backend:
 Now we create a repository.
 
 .. jupyter-execute::
+    :hide-output:
 
     audbackend.create('sql', './host', 'repo')
 
--- a/docs/developer-guide.rst
+++ b/docs/developer-guide.rst
@@ -155,7 +155,8 @@ and upload a file:
 
     import audeer
 
-    interface = audbackend.create('file-system', './host', 'repo', interface=UserContent)
+    audbackend.create('file-system', './host', 'repo')
+    interface = audbackend.access('file-system', './host', 'repo', interface=UserContent)
 
     interface.add_user('audeering', 'pa$$word')
     audeer.touch('local.txt')
@@ -320,11 +321,11 @@ stored on our backend:
             db.execute(query)
 
 
-Now we create an instance.
+Now we create a repository.
 
 .. jupyter-execute::
 
-    interface = audbackend.create('sql', './host', 'repo')
+    audbackend.create('sql', './host', 'repo')
 
 
 We also add a method to access
--- a/docs/developer-guide.rst
+++ b/docs/developer-guide.rst
@@ -1,0 +1,626 @@
+.. set temporal working directory
+.. jupyter-execute::
+    :hide-code:
+
+    import os
+    import audeer
+
+    _cwd_root = os.getcwd()
+    _tmp_root = audeer.mkdir(os.path.join('docs', 'tmp-developer-guide'))
+    os.chdir(_tmp_root)
+
+
+.. _developer-guide:
+
+Developer guide
+===============
+
+The aim of
+:mod:`audbackend`
+is to provide an
+abstract interface for
+any kind of file storage system.
+Even those,
+that have not been
+invented yet :)
+
+This tutorial is divided
+into three parts.
+Under :ref:`register-a-backend`,
+we show how existing or new backends
+are assigned to names
+used when creating a repository with
+:func:`audbackend.create`
+or when accessing a repository with
+:func:`audbackend.access`.
+In :ref:`develop-new-interface`
+we show how to create an interface
+that manages user content.
+Under :ref:`develop-new-backend`,
+we take a deep dive
+and develop a backend
+that stores files into
+a SQLite_ database.
+
+
+.. _register-a-backend:
+
+Register a backend
+------------------
+
+Backends are referred to by names.
+The current active names can be shown by:
+
+.. jupyter-execute::
+
+    import audbackend
+
+    list(audbackend.available())
+
+We can assign a name to a backend
+by registering a backend class.
+
+.. jupyter-execute::
+
+    audbackend.register('files', audbackend.backend.FileSystem)
+    list(audbackend.available())
+
+
+.. _develop-new-interface:
+
+Develop new interface
+---------------------
+
+We can implement our own interface
+by deriving from
+:class:`audbackend.interface.Base`.
+For instance,
+we can create an interface
+to manage user content.
+It provides three functions:
+
+* ``add_user()`` to register a user
+* ``upload()`` to upload a file for user
+* ``ls()`` to list the files of a user
+
+We store user information
+in a database under
+``'/user.map'``.
+To access and update
+the database
+we implement the following
+helper class.
+
+
+.. jupyter-execute::
+
+    import shelve
+
+    class UserDB:
+        r"""User database.
+
+        Temporarily get user database
+        and write changes back to the backend.
+
+        """
+        def __init__(self, backend: audbackend.backend.Base):
+            self.backend = backend
+
+        def __enter__(self) -> shelve.Shelf:
+            if self.backend.exists('/user.db'):
+                self.backend.get_file('/user.db', '~.db')
+                self._map = shelve.open('~.db', flag='w', writeback=True)
+            else:
+                self._map = shelve.open('~.db', writeback=True)
+            return self._map
+
+        def __exit__(self, exc_type, exc_val, exc_tb):
+            self._map.close()
+            self.backend.put_file('~.db', '/user.db')
+            os.remove('~.db')
+
+
+Now,
+we implement the interface.
+
+.. jupyter-execute::
+
+    class UserContent(audbackend.interface.Base):
+
+        def add_user(self, username: str, password: str):
+            r"""Add user to database."""
+            with UserDB(self.backend) as map:
+                map[username] = password
+
+        def upload(self, username: str, password: str, path: str):
+            r"""Upload user file."""
+            with UserDB(self.backend) as map:
+                if username not in map or map[username] != password:
+                    raise ValueError('User does not exist or wrong password.')
+                self.backend.put_file(path, f'/{username}/{os.path.basename(path)}')
+
+        def ls(self, username: str) -> list:
+            r"""List files of user."""
+            with UserDB(self.backend) as map:
+                if username not in map:
+                    return []
+            return self.backend.ls(f'/{username}/')
+
+
+Let's create a backend
+with our custom interface:
+
+.. jupyter-execute::
+
+    import audeer
+
+    backend = audbackend.create('file-system', './host', 'repo', interface=UserContent)
+
+    backend.add_user('audeering', 'pa$$word')
+    audeer.touch('local.txt')
+    backend.upload('audeering', 'pa$$word', 'local.txt')
+    backend.ls('audeering')
+
+
+At the end we clean up and delete our repo.
+
+.. jupyter-execute::
+
+    audbackend.delete('file-system', './host', 'repo')
+
+
+.. _develop-new-backend:
+
+Develop new backend
+-------------------
+
+In the previous section
+we have used an existing
+backend implementation.
+Now we develop a new backend
+that implements
+a SQLite_ database.
+
+A new backend
+should be implemented as a class
+deriving from
+:class:`audbackend.backend.Base`.
+As can be seen in the file
+:file:`audbackend/core/backend/base.py`,
+we need to implement the following private methods:
+
+* ``_access()``
+* ``_checksum()``
+* ``_create()``
+* ``_date()``
+* ``_delete()``
+* ``_exists()``
+* ``_get_file()``
+* ``_ls()``
+* ``_owner()``
+* ``_put_file()``
+* ``_remove_file()``
+
+We call the class ``SQLite``.
+and we add two more attributes
+in the constructor:
+
+* ``_path``: the path of the database,
+  which we derive from the host and repository,
+  namely ``'<host>/<repository>/db'``.
+* ``_db``: connection object to the database.
+
+.. jupyter-execute::
+
+    import audbackend
+    import os
+
+    class SQLite(audbackend.backend.Base):
+
+        def __init__(
+                self,
+                host: str,
+                repository: str,
+        ):
+            super().__init__(host, repository)
+            self._path = os.path.join(host, repository, 'db')
+            self._db = None
+
+
+Obviously,
+this is not yet a fully
+functional backend implementation.
+But for the sake of clarity,
+we will dynamically add
+the required methods one after another
+using a dedicated decorator:
+
+.. jupyter-execute::
+
+    import functools
+
+    def add_method(cls):
+        def decorator(func):
+            @functools.wraps(func)
+            def wrapper(self, *args, **kwargs):
+                return func(self, *args, **kwargs)
+            setattr(cls, func.__name__, wrapper)
+            return func
+        return decorator
+
+For instance,
+to ensure the connection to the database
+is properly closed,
+we add a destructor method.
+This is not mandatory
+and whether it is needed
+depends on the backend.
+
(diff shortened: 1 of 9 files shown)


hagenw:
There is still one remaining problem with the "Usage" documentation. We always use `backend` as variable name. Which might be misleading as a user might expect she/he needs to look under the API of the used backend class instead of the API of the used interface class. Should we rename the variable to `interface`, `storage`, or `repo`; or should we add an extra paragraph explaining which API needs to be looked at?

frankenjoe:
> Should we rename the variable to interface

I had the same thought. I think it makes sense to rename accordingly.

> When using audbackend.create() or audbackend.access() the backend is referred to by a string name whereas the interface needs to be given as a class. Providing it as a class makes things easier, but as a user you might be wonder why there is this difference.

I would argue that for a specifc repository, we want to share exactly one backend object. Whereas we can have multiple interface objects to operate on the same backend.

> audbackend.available() does only list backends, not interfaces

I guess we can argue similar as before. It's only the backends that are unique, whereas we do not keep track of the interfaces.

> audbackend.available() can only provide information on the backend class associated with a backend name, once an instance of the backend is created.

I think that is the purpose of `available()`: list all repositories I can currently access.

---

But in general I agree of course, that the current API is sub-optimal. More logical would be:

```python
backend = audbackend.create(...)  # either work directly with backend object if only unversiond access is required
interface = audackend.interface.Versioned(backend)   # or wrap with an interface object, e.g. if a versioned access is required
```

But since `backend` provides unversioned access now, this would be breaking change. That's why I decided to add that interface argument and set it to `audbackend.interface.Versioned`.

If we really want to get rid of providing interface as a class, we add a `audbackend.register_interface()` function:

```python
audbackend.register('file-system', audbackend.backend.FileSystem)
audbackend.register_interface('unversioned', audbackend.interface.Unversioned)   # register `audbackend.interface.Versioned` by default as `'versioned'`
interface = audbackend.create('file-system', './host', 'repo', interface='unversioned')  # defaults to 'versioned'
```

But I am not really sure if this is really needed.

hagenw:
> > Should we rename the variable to interface
> 
> I had the same thought. I think it makes sense to rename accordingly.

I renamed the variable to `interface`.

hagenw:
> > When using audbackend.create() or audbackend.access() the backend is referred to by a string name whereas the interface needs to be given as a class. Providing it as a class makes things easier, but as a user you might be wonder why there is this difference.
> 
> I would argue that for a specifc repository, we want to share exactly one backend object. Whereas we can have multiple interface objects to operate on the same backend.
> 
> > audbackend.available() does only list backends, not interfaces
> 
> I guess we can argue similar as before. It's only the backends that are unique, whereas we do not keep track of the interfaces.
> 
> > audbackend.available() can only provide information on the backend class associated with a backend name, once an instance of the backend is created.
> 
> I think that is the purpose of `available()`: list all repositories I can currently access.
> 
> But in general I agree of course, that the current API is sub-optimal. More logical would be:
> 
> ```python
> backend = audbackend.create(...)  # either work directly with backend object if only unversiond access is required
> interface = audackend.interface.Versioned(backend)   # or wrap with an interface object, e.g. if a versioned access is required
> ```
> 
> But since `backend` provides unversioned access now, this would be breaking change. That's why I decided to add that interface argument and set to `audbackend.interface.Versioned` by default.
> 
> If we really want to get rid of providing interface as a class, we add a `audbackend.register_interface()` function:
> 
> ```python
> audbackend.register('file-system', audbackend.backend.FileSystem)
> audbackend.register_interface('unversioned', audbackend.interface.Unversioned)   # register `audbackend.interface.Versioned` by default as `'versioned'`
> interface = audbackend.create('file-system', './host', 'repo', interface='unversioned')  # defaults to 'versioned'
> ```
> 
> But I am not really sure if this is really needed.

No, I would stay with the current solution.

frankenjoe:
What still bothers me most is that we we have the new `interface` argument also with `create()`. This implies that the created backend is somehow bound to the interface, which is not the case. To avoid a breaking change we could maybe remove the argument `create()` and return a versioned interface but not document it.

hagenw:
> What still bothers me most is that we we have the new `interface` argument also with `create()`. This implies that the created backend is somehow bound to the interface, which is not the case. To avoid a breaking change we could maybe remove the argument from `create()` and silently return a versioned interface that we not document.

I would not remove the argument, as it returns an interface and not a backend. It feels strange to me, that I cannot decide which interface I would like to have. Another solution would be, that `create()` does not return anything, which of cause would also be a breaking change.

frankenjoe:
What I suggest is that it silently returns a versioned interface to avoid a breaking change, but that we do not document it.

hagenw:
> What I suggest is that it silently returns a versioned interface to avoid a breaking change, but that we do not document it.

But to me it makes no sense, when I want to have a unversioned interface, why I shouldn't be able to request this already from `create()`? If you don't provide the interface argument, you will get the default interface, which is fine.

frankenjoe:
Well, in pratice you will rarely write code where you call `create()` and do something with the returned interface, unless you delete the repository afterwards (e.g. in a test). Since the code will fail when you run it again. So the common use-case is that you have `create()` called once in an init script or something. And then you have the code where you actually interact with the backend and where you call `access()`. So to me it makes sense to not have a return value for `create()`.

hagenw:
> So to me it makes sense to not have a return value for `create()`.

I agree. I created https://github.com/audeering/audbackend/issues/173to tackle this independent of this pull request.

frankenjoe:
I suggest we first solve #173 before we continue here.

hagenw:
Yes, good idea.

frankenjoe:
Will you tackle https://github.com/audeering/audbackend/issues/175 in this PR or would you prefer we do it in a separate PR? I can also take care of it if you like.

hagenw:
Please go ahead in another PR, I will focus on the usage documentation here.

hagenw:
I have now updated the docs reflecting the changes to `create()` introduced in #173

docs/usage.rst
```diff
@@ -33,180 +6,152 @@ a SQLite_ database.
...
-by default,
-but it doesn't hurt
-if we do it again).
+The user can select between different :ref:`interfaces <interfaces>`
+that influence how the data is structured
+and presented to the user.
+In addition,
+:mod:`audbackend` supports different storage systems,
+so called :ref:`backends <backends>`.
+The names of all available backends can be listed by:
```
frankenjoe, line 29:
```suggestion
The names of all available backends can be listed with:
```

frankenjoe: At least sounds better to me, but I am not 100% sure :)

docs/usage.rst
```diff
@@ -33,180 +6,152 @@ a SQLite_ database.
...
-* ``name``: the name under which the backend class is registered
-* ``host``: the host address,
-  in this case a folder on the local file system.
-* ``repository``: the repository name,
-  in this case a sub-folder within the ``host`` folder
-  (it is possible to have several repositories
-  on the same host).
+As all data is stored in a repository
+on the backend,
+we need to first create one.
```
frankenjoe, line 45:
```suggestion
To store data on a backend
we need to create a repository first.
```

docs/usage.rst
```diff
@@ -33,180 +6,152 @@ a SQLite_ database.
...
-implements
-:class:`audbackend.backend.Base`
-for a standard file system.
+Usage
+=====
 
+With the help of :mod:`audbackend`
+a user can store files
+in a repository
+on a storage system.
```
frankenjoe, line 21:
```suggestion
on a storage system
(backend).
```

docs/usage.rst
```diff
@@ -33,180 +6,152 @@ a SQLite_ database.
...
 
-We start by registering the class
-(in fact,
-the class is registered
-by default,
-but it doesn't hurt
-if we do it again).
+The user can select between different :ref:`interfaces <interfaces>`
+that influence how the data is structured
+and presented to the user.
```
frankenjoe, line 25:
```suggestion
File access is handled
via an :ref:`interface <interfaces>`,
which defines how the data is structured
and presented to the user.
```

docs/usage.rst
```diff
@@ -33,180 +6,152 @@ a SQLite_ database.
...
-    audbackend.available()
-
-
-We can access an existing repository with:
+Once we have an existing repository,
+we can access it with :func:`audbackend.access`.
+We choose the :class:`audbackend.interface.Unversioned` interface
+to communicate with the repository,
+instead of the default one
+(:class:`audbackend.interface.Versioned`).
```
frankenjoe, line 58: Is it important to mention the default here? Though it's probably obvious I think we should explain what unversioned access means, e.g. "We use the Unversioned interface. It does not support versioning, i.e. exactly one file exists for a backend path."

docs/usage.rst
```diff
@@ -276,584 +196,68 @@ exception thrown by the backend.
...
-        './host',
-        'repo',
-        interface=audbackend.interface.Unversioned,
-    )
-    backend.put_file('local.txt', '/file-without-version.txt')
-    backend.ls()
+We start by creating a repository
+on the ``'file-system'`` backend
+and accessing it
+using the default :class:`audbackend.interface.Versioned` interface.
```
frankenjoe, line 208:
```suggestion
on the ``'file-system'`` backend.
This time we access it
with the :class:`audbackend.interface.Versioned` interface
(which is also used by default).
```

docs/usage.rst
```diff
@@ -276,584 +196,68 @@ exception thrown by the backend.
...
-
-    backend.exists('/file.txt', '1.0.0')
-
-
-And a method that uploads
-a file to our backend.
-
-.. jupyter-execute::
+    audbackend.create('file-system', './host', 'repo')
+    interface = audbackend.access('file-system', './host', 'repo')
```
frankenjoe, line 213:
```suggestion
    interface = audbackend.access(
        'file-system',
        './host',
        'repo',
        interface=audbackend.interface.Versioned,
    )
```

I would be verbose here.

docs/usage.rst
```diff
@@ -276,584 +196,68 @@ exception thrown by the backend.
...
-            '''
-            owner = getpass.getuser()
-            date = datetime.datetime.today().strftime('%Y-%m-%d')
-            data = (dst_path, checksum, content, date, owner)
-            db.execute(query, data)
-
-
-Let's put a file on the backend.
+We then upload a file and assign it ``'1.0.0'``
+as its version.
```
frankenjoe, line 216:
```suggestion
We then upload a file and 
assign version ``'1.0.0'`` to it.
```

docs/usage.rst
```diff
@@ -276,584 +196,68 @@ exception thrown by the backend.
...
-                FROM data
-                WHERE path="{path}"
-            '''
-            date = db.execute(query).fetchone()[0]
-        return date
-
-    backend.date('/file.txt', '1.0.0')
+If we inspect the content of the repository
+it will return a list of tuples
+containing name and version.
```
frankenjoe, line 234:
```suggestion
containing file name and version.
```

docs/developer-guide.rst
```diff
@@ -0,0 +1,629 @@
...
+Even those,
+that have not been
+invented yet :)
+
+This tutorial is divided
+into three parts.
+Under :ref:`register-a-backend`,
+we show how existing or new backends
+are assigned to names,
+that are used by
```
frankenjoe, line 32:
```suggestion
we show how a backend class
is assigned to an alias name
that is used to create and access
repositories with
```

docs/developer-guide.rst
```diff
@@ -0,0 +1,629 @@
...
+:func:`audbackend.create`
+and
+:func:`audbackend.access`.
+In :ref:`develop-new-interface`,
+we show how to create an interface
```
frankenjoe, line 37:
```suggestion
we show how to create a custom interface
```

docs/developer-guide.rst
```diff
@@ -0,0 +1,629 @@
...
+a SQLite_ database.
+
+
+.. _register-a-backend:
+
+Register a backend
+------------------
+
+Backends are referred to by names.
+The current active names can be listed by:
```
frankenjoe, line 52:
```suggestion
A backend can host
an arbitrary number of repositories.
Repositories created or accessed
in the current session
can be listed with:
```

frankenjoe: I am not sure if "session" is the best term here...

docs/developer-guide.rst
```diff
@@ -0,0 +1,629 @@
...
+
+.. jupyter-execute::
+
+    import audbackend
+
+    list(audbackend.available())
+
+We can assign a name to a backend
+by registering its backend class.
```
frankenjoe, line 61:
```suggestion
We register a backend class
by assigning an alias to it.
Functions like
:func:`audbackend.create`
and
:func:`audbackend.access`
expect the alias instead
of the class name.
This makes it easier
to read available repositories
from a file.
```

docs/usage.rst
hagenw: I also had first "with", but it sounded to German to me ;)

I searched a littel bit and "listed by" seems to be used to refer to persons oding the listing, whereas "listed with" refers to objects that are listed together with the object before. So maybe "by" is then better as the code that follows is responsible for the listing and is not listed together with the text before?

docs/developer-guide.rst
hagenw: Interesting, this fits into our discussion in https://github.com/audeering/audbackend/pull/176#discussion_r1467692547.

`audbackend.available()` is of cause very blind to already existing repositories. E.g. you can get back an empty list, but you might still get an error when ying to create a repo, but then it turns out the repo exists already.
And on the other hand, a repo becomes part of the list returned by `audbackend.available()` after you have accessed it. But then the information you get from `audbackend.available()` is zero.

But yes, for the current behavior of `audbackend.available()` your wording seems to be fine with me.

docs/developer-guide.rst
hagenw: The last sentence I don't get. You mean it is easier to define repositories in config files, correct?

docs/usage.rst
hagenw: I adjusted the text accordingly.

docs/usage.rst
hagenw: I changed to:

```
We then upload a file
and assign version ``'1.0.0'`` to it.
```

docs/usage.rst
frankenjoe: Ok, let's stick with "by" then :)

docs/developer-guide.rst
frankenjoe: Yes, I think that was actually the motivation why we introduced that whole concept of alias names. In the end, we need to keep track of the repositories we e.g. want to search for databases in case of audb.

hagenw: I applied your suggested changes, and added `(config)` in front of `file`.

frankenjoe:
Could you please update the images in the description or have you already done so?

PR #172, opened 2024-01-23
Backend sub module

frankenjoe:


hagenw:
Great, the separation is much nicer now between interface and backends.
It's also nice to have the backward compatibility with `audbackend.Backend`, but I guess the bigger chnage is anyway that `audbackend.create()` and `audbackend.access()` now return iinterfaces and not backends. But I guess it will be mainly the type hints that we need to fix in  other packages.

PR #174, opened 2024-01-25
create(): remove return value

### LINKED ISSUE #173: Change `audbackend.create()` to not return an interface
As discussed in https://github.com/audeering/audbackend/pull/171#issuecomment-1909715067 and https://github.com/audeering/audbackend/pull/171#issuecomment-1909779427 it makes sense that `audbackend.create()` does not return an interface as the interface should be independent of the backend.

This would have a few other consequences:
* we don't have to think about how to store the interface belonging to a repo as part of the repo (as proposed in https://github.com/audeering/audbackend/issues/166)
* we might need to add some error handling for the case that two interafces access the same repo that are not compatible with each other, e.g. broken `ls()` methods or something similar

frankenjoe:
Closes #173 

This (officially) removes the return value from `audbackend.create()`. For legacy reasons we still return an instance of `audbackend.interface.Versioned`.

image

### DIFF
--- a/audbackend/core/api.py
+++ b/audbackend/core/api.py
@@ -133,16 +133,12 @@ def create(
         name: str,
         host: str,
         repository: str,
-        *,
-        interface: typing.Type[Interface] = Versioned,
-        interface_kwargs: dict = None
-) -> Interface:
+ ):
     r"""Create repository.
 
     Creates ``repository`` on the ``host``
-    and returns an ``interface`` instance for it.
-    The backend is an object of the class
-    registered under the alias ``name``
+    using the class registered
+    under the alias ``name``
     with :func:`audbackend.register`.
 
     If the repository cannot be created
@@ -159,11 +155,6 @@ def create(
         name: alias under which backend class is registered
         host: host address
         repository: repository name
-        interface: interface class
-        interface_kwargs: keyword arguments for interface class
-
-    Returns:
-        interface object
 
     Raises:
         BackendError: if an error is raised on the backend,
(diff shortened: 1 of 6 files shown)


audbackend/core/api.py
```diff
@@ -133,16 +133,12 @@ def create(
...
-        interface_kwargs: dict = None
-) -> Interface:
+ ):
     r"""Create repository.
 
     Creates ``repository`` on the ``host``
-    and returns an ``interface`` instance for it.
-    The backend is an object of the class
-    registered under the alias ``name``
+    using the class registered
```
hagenw, line 140:
```suggestion
    using the backend class registered
```

audbackend/core/api.py
```diff
@@ -133,16 +133,12 @@ def create(
         name: str,
         host: str,
         repository: str,
-        *,
-        interface: typing.Type[Interface] = Versioned,
-        interface_kwargs: dict = None
-) -> Interface:
+ ):
```
hagenw, line 136:
```suggestion
):
```

hagenw:
In the text of the docstring the `BackendError` is also discussed. I think we can remove the paragraph as it is repeated in the `Raises` section.

frankenjoe:
> In the text of the docstring the `BackendError` is also discussed. I think we can remove the paragraph as it is repeated in the `Raises` section.

Good point. Changed it also for `access()`.

hagenw:
Cool, thanks.

Should we maybe add a note to the docstring of `audbackend.create()` stating that it still returns an interface object. but that it will be removed in a future version. Or should we stay silent about it as it's also not that important?

frankenjoe:
> Should we maybe add a note to the docstring of `audbackend.create()` stating that it still returns an interface object. but that it will be removed in a future version. Or should we stay silent about it as it's also not that important?

Yes, I guess a note would not hurt. One problem is of course that we cannot show a deprecation warning, since we do not know when the user uses the returned value. So maybe we simply say it might be removed in a future version?

And one more thing I noticed. Lately we started to talk about "repository" instead of "backend object" (see description). To me it makes more sense since the common user will not directly interact with the backend object anyway and repository is probably more descriptive. But maybe we should unify it across all api functions. In `available()` we still talk about backend objects, for instance. Maybe something to keep in mind in #171.

hagenw:
Yes, I also switched talking more about repository in #171 (but so far restricted to the usage section of the documentation).

I would simply write future version in the notes as it is not so easy to forsee the upcoming versions of `audbackend` ;)

frankenjoe:
I added the following note:

image

PR #176, opened 2024-01-26
DOC: use repository over backend object in all api functions

### LINKED ISSUE #175: DOC: Use term 'repository' over 'backend instance' in API
As mentioned in https://github.com/audeering/audbackend/pull/174#issuecomment-1911640753 we should use the term `'repository'` over `'backend instance'` in the remaining API functions, namely `access()`, `available()` and `delete()`.

frankenjoe:
Closes #175 

image

image

image

image

image

### DIFF
--- a/audbackend/core/api.py
+++ b/audbackend/core/api.py
@@ -56,16 +56,17 @@ def access(
 
     Returns an ``interface`` instance
     to access the ``repository``
-    on the ``host``.
-    The backend is an object of the class
-    registered under the alias ``name``
-    with :func:`audbackend.register`.
+    located at ``host``
+    on the backend with alias ``name``
+    (see :func:`audbackend.register`).
 
     Use :func:`audbackend.available`
-    to list available backend instances.
+    to list available repositories
+    or :func:`audbackend.create`
+    to create a new repository.
 
     Args:
-        name: alias under which backend class is registered
+        name: backend alias
         host: host address
         repository: repository name
         interface: interface class
@@ -92,16 +93,17 @@ def access(
 
 
 def available() -> typing.Dict[str, typing.List[Base]]:
-    r"""List available backend instances.
+    r"""List available repositories.
 
     Returns a dictionary with
-    registered backend aliases as keys
+    backend alias name as key
     (see :func:`audbackend.register`)
-    and a list with backend instances as values
+    and a list with repositories
+    on that backend as value
     (see :func:`audbackend.create`).
 
     Returns:
-        dictionary with backend instances
+        dictionary with repositories
 
     Examples:
         >>> list(available())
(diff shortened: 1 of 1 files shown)


audbackend/core/api.py
```diff
@@ -92,16 +93,17 @@ def access(
...
-    r"""List available backend instances.
+    r"""List available repositories.
 
     Returns a dictionary with
-    registered backend aliases as keys
+    backend alias name as key
     (see :func:`audbackend.register`)
-    and a list with backend instances as values
+    and a list with repositories
+    on that backend as value
```
hagenw, line 102: Not exactly.
At the moment it returns a list of tuples containing backend class, host path, repo name:
```python
>>> import audbackend
>>> audbackend.create("file-system", ".", "repo")
>>> audbackend.available()["file-system"]
[('audbackend.core.backend.filesystem.FileSystem', '.', 'repo')]
```

A list with repositories would look like this to me:
```python
>>> audbackend.available()["file-system"]
[Repository('repo', '.', 'file-system')]
```

audbackend/core/api.py
```diff
@@ -92,16 +93,17 @@ def access(
 
 
 def available() -> typing.Dict[str, typing.List[Base]]:
```
hagenw, line 95: The type hint for the return value looks wrong to me, it should be:

```python
typing.Dict[str, typing.List[typing.Tuple[Base, str, str]]]
```

frankenjoe: I think it's correct, it looks like this because of the way we implemented `Backend.__repr__()`:

```python
backend = audbackend.Backend('host', 'repo')
repr(backend)
```
```
('audbackend.core.backend.base.Base', 'host', 'repo')
``

hagenw: Ok, I guess the confusion comes from the fact that we actually have two classes that represent an repository on a backed: `audbackend.backend.Base` which needs to be instanciated with `host` and `repo`; and we have `audbackend.Repository` which needs to instanciated with `name` (of the repository), `host` and `backend` (but name not class).

hagenw: In other packages that use `audbackend` we then take the `audbackend.Repository` class when presenting the concept of a repository to the user, e.g. `audb.Repository`.

frankenjoe: Yes, I think we actually introduced `audbackend.Repository` when we found such a concept was useful in `audb`. I am not a 100% sure though.

audbackend/core/api.py
hagenw: For me it is still not completely clear what should be the purpose of `audbackend.available()`.
If it should list available repsoitories it shouldn't return a dictionary, but simply a list of repository objects as those state anyway which abckend they are.

I think we introduced its current behavior to not only show us existing repositories, but also which backend alias names are linked to which backend classes. Otherwise I see no use at all of the returned dict.

frankenjoe: Right, the keys list all registered names, even for backends where have no "active" repositories.

PR #178, opened 2024-01-31
Base._ls(): may always return an empty list

frankenjoe:
So far the docstring of `Base._ls()` was saying:

```
        If ``path`` is ``'/'`` and no files exist on the repository,
        an empty list should be returned
        Otherwise,
        if ``path`` does not exist or no files are found under ``path``,
        an error should be raised.
```

This might be overlooked by a developer, so here we simplify `_ls()` by not demanding to raise an error message or treat `/` as a special case. Instead it is sufficient to return an empty list, even in the case that the path does not exist. The docstring now says:

```
        If ``path`` does not exist
        an empty list can be returned.
```

To avoid a breaking change, raising a file-not-found-error is now handled by `Base`, in the same way we already do it for the case that `path` is not a sub-path.

### DIFF
--- a/audbackend/core/backend/base.py
+++ b/audbackend/core/backend/base.py
@@ -348,11 +348,9 @@ class Base:
     ) -> typing.List[str]:  # pragma: no cover
         r"""List all files under sub-path.
 
-        If ``path`` is ``'/'`` and no files exist on the repository,
-        an empty list should be returned
-        Otherwise,
-        if ``path`` does not exist or no files are found under ``path``,
-        an error should be raised.
+        If ``path`` does not exist
+        or no files are found under ``path``
+        return an empty list.
 
         """
         raise NotImplementedError()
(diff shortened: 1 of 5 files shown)


frankenjoe:
@hagenw locally tests with Artifactory are passing, but here it seems that it can no longer access the server. Any idea?

hagenw:
Maybe there is also an error with the stored secret on Github. I will try to replace it with a new one.

frankenjoe:
But isn't it strange we only encounter this problem now?

frankenjoe:
> Maybe there is also an error with the stored secret on Github. I will try to replace it with a new one.

Ok, thanks.

hagenw:
> But isn't it strange we only encounter this problem now?

I also encountered the problem when I tried to add tests for non-existing and private repositories in https://github.com/audeering/audb/pull/339

hagenw:
I replaced the secret on Github and re-started the CI jobs, but unfortunately the error stays the same.

hagenw:
It seems to be common that secrets are replaced by "***" in the output of CI jobs. Here is the output of an older passing one:

image

hagenw:
I tested a few more things and the username and API key seem still to be correctly set, but the error remains in the end:

```
E           dohq_artifactory.exception.ArtifactoryException: This request is blocked due to recurrent login failures, please try again in 13 seconds
```

Maybe we wait until next week to see if the error remains?

hagenw:
As usual, waiting a few days helps to solve login issues.

audbackend/core/backend/base.py
```diff
@@ -419,17 +416,19 @@ def ls(
...
-                    except FileNotFoundError as ex:
-                        raise BackendError(ex)
                 paths = []
 
         if not paths:
+
+            if path != '/' and not suppress_backend_errors:
+                # since the backend does no longer raise an error
+                # if the path does not exist
+                # we have to do it
```
hagenw, line 426: The first part of the comment will not make much sense in the future, so maybe change to:

```suggestion
                # if the path does not exist
                # we raise an error
```

If I remember correctly, this behavior makes sense as we cannot have empty sub-paths on the backend, correct?

PR #179, opened 2024-02-01
DEBUG authentication error in CI

hagenw:
Debug authentication errors we experience in https://github.com/audeering/audbackend/pull/178

### DIFF
--- a/tests/test_backend_artifactory.py
+++ b/tests/test_backend_artifactory.py
@@ -33,6 +33,18 @@ def hide_credentials():
             del os.environ[key]
 
 
+def test_env_variables():
+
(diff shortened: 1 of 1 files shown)


hagenw:
@frankenjoe I tried to debug the content of the username and api key env variables here by printing them to stdout.
Thios is what I get:

```
----------------------------- Captured stdout call -----------------------------
140
os.environ['ARTIFACTORY_USERNAME']='***'
141
os.environ['ARTIFACTORY_API_KEY']='***'
```

So I'm not sure yet if Github somehow manages to avoid them being displayed or if our current approach how we assign those variables no longer works.

hagenw:
No longer needed.

The authentication issue was temporary and went away after waiting a few days.

Issue #177, opened 2024-01-26
Add a move / copy function


frankenjoe:
It would be nice to provide a function that moves / copies a file on a backend. For both operations we can provide a default implementation where a file is temporarily downloaded, optionally removed from the backend and uploaded again under the target path. However, since some backends may support moving / copying as an atomic operation on the backend, we should also allow an backend to overwrite the default behaviour.

PR #181, opened 2024-02-07
Copy file on backend

frankenjoe:
Relates to #177 

Adds `backend.Base.copy_file()` to copy a file on the backend. A default implementation is provided, which downloads the files to a temporary directory and afterward uploads it to the new location. The behavior can be overwritten by implementing a custom `_copy_file()` method. In case of a versioned interface, all versions of file are copied unless a specific version is selected.

#### backend.Base / backend.Artifactory / backend.FileSystem

image

#### interface.Unversioned

image

#### interface.Versioned

image

### Usage

image

...

image

### Developer

image

### DIFF
--- a/audbackend/core/backend/base.py
+++ b/audbackend/core/backend/base.py
@@ -71,6 +71,70 @@ class Base:
             path,
         )
 
+    def _copy_file(
+            self,
+            src_path: str,
+            dst_path: str,
+            verbose: bool,
+    ):
+        r"""Copy file on backend.
+
+        A default implementation is provided,
+        which temporarily gets the file from the backend
+        and afterward puts it to the new location.
+        It is recommended to overwrite the function
+        if backend supports a native way to copy files.
+
+        """
+        with tempfile.TemporaryDirectory() as tmp:
+            tmp_path = audeer.path(tmp, '~')
+            tmp_path = self.get_file(src_path, tmp_path, verbose=verbose)
+            self.put_file(tmp_path, dst_path, verbose=verbose)
+
+    def copy_file(
+            self,
+            src_path: str,
+            dst_path: str,
+            *,
+            verbose: bool = False,
+    ):
+        r"""Copy file on backend.
+
+        If ``dst_path`` exists
+        with a different checksum,
+        it is overwritten,
+        or otherwise,
+        the operation is silently skipped.
+
+        Args:
+            src_path: source path to file on backend
+            dst_path: destination path to file on backend
+            verbose: show debug messages
+
+        Raises:
+            BackendError: if an error is raised on the backend
+            ValueError: if ``src_path`` or ``dst_path``
+                does not start with ``'/'`` or
+                does not match ``'[A-Za-z0-9/._-]+'``
+
+        """
+        src_path = utils.check_path(src_path)
+        dst_path = utils.check_path(dst_path)
+
+        if (
+            src_path != dst_path
+            and (
+                not self.exists(dst_path)
+                or self.checksum(src_path) != self.checksum(src_path)
+            )
+        ):
+            utils.call_function_on_backend(
+                self._copy_file,
+                src_path,
+                dst_path,
+                verbose,
+            )
+
     def _create(
             self,
     ):  # pragma: no cover
@@ -259,7 +323,7 @@ class Base:
 
         To ensure the file is completely retrieved,
         it is first stored in a temporary directory
-        and afterwards moved to ``dst_path``.
+        and afterward moved to ``dst_path``.
 
         Args:
             src_path: path to file on backend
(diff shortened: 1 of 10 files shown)


audbackend/core/backend/base.py
```diff
@@ -71,6 +71,70 @@ def checksum(
             path,
         )
 
+    def _copy_file(
+            self,
+            src_path: str,
+            dst_path: str,
+            verbose: bool,
+    ):  # pragma: no Windows cover
```
frankenjoe, line 79: We need this because `FileSystem` and `Artifactory' provide a custom implementation and `SingleFolder`, which relies on the default implementation, is not tested on Windows.

audbackend/core/backend/base.py
```diff
@@ -71,6 +71,70 @@ def checksum(
...
+            *,
+            verbose: bool = False,
+    ):
+        r"""Copy file on backend.
+
+        If ``dst_path`` exists
+        with a different checksum,
+        it is overwritten,
+        or otherwise,
+        the operation is silently skipped.
```
hagenw, line 107: Maybe change to

```suggestion
        If ``dst_path`` exists
        and has a different checksum,
        it is overwritten.
        Otherwise,
        the operation is silently skipped.
```

audbackend/core/interface/unversioned.py
```diff
@@ -37,6 +37,42 @@ def checksum(
...
+            *,
+            verbose: bool = False,
+    ):
+        r"""Copy file on backend.
+
+        If ``dst_path`` exists
+        with a different checksum,
+        it is overwritten,
+        or otherwise,
+        the operation is silently skipped.
```
hagenw, line 53: As above:

```suggestion
        If ``dst_path`` exists
        and has a different checksum,
        it is overwritten.
        Otherwise,
        the operation is silently skipped.
```

hagenw: BTW, what happens if a sub path exists on the backend with the same name as the file. I guess trying to overwrite it will raise an error?
Would be nice to add a test for this case.

audbackend/core/interface/versioned.py
```diff
@@ -59,6 +60,62 @@ def checksum(
...
+
+        If ``version`` is ``None``
+        all versions of ``src_path``
+        will be copied.
+
+        If ``dst_path`` exists
+        with a different checksum,
+        it is overwritten,
+        or otherwise,
+        the operation is silently skipped.
```
hagenw, line 81: As above:

```suggestion
        If ``dst_path`` exists
        and has a different checksum,
        it is overwritten.
        Otherwise,
        the operation is silently skipped.
```

docs/usage.rst
```diff
@@ -256,6 +255,20 @@ Or request it's latest version.
 
     interface.latest_version('/file.txt')
 
+When can copy a specific version of a file.
```
hagenw, line 258:
```suggestion
We can copy a specific version of a file.
```

tests/singlefolder.py
```diff
@@ -153,7 +153,7 @@ def _put_file(
     ):
         with self.Map(self._path, self._lock) as m:
 
-            if dst_path not in m:
+            if dst_path not in m or checksum != m[dst_path][1]:
```
hagenw, line 156: Why do we need this extra check here?

frankenjoe: Because we only want to skip when the checksums match. Otherwise the file has changed and we need to replace it. Was a bug before.

audbackend/core/backend/base.py
hagenw: It's not ideal to skip testing the default implementation on Windows.
What was the reason that `SingleFolder` is not tested under Windows?

frankenjoe:
> The case that it might fail is covered by the `BackendError` in the docstring then I guess?

I was expecting so, but turns out it does not :) This is because `FileSystem` uses `shutil.copy()` under the hood, which inteprets `/sub` as a folder if it exists and in that case create a file with the name of the source path in that folder. Here is a minimal example that reveals this behavior:

```python
import audeer

import audbackend

with tempfile.TemporaryDirectory() as tmp:

    path = audeer.touch(tmp, 'file.txt')
    backend = audbackend.backend.FileSystem(tmp, 'repo')
    backend.put_file(path, '/sub/file.txt')
    path = audeer.touch(tmp, 'file2.txt')
    backend.put_file(path, '/sub')
    backend.ls()
```
```
['/sub/file.txt', '/sub/file2.txt']
```

frankenjoe:
But I suggest we discuss this in a separate issue, as it is not directly related to this PR.

hagenw:
OK, I created https://github.com/audeering/audbackend/issues/184.

PR #185, opened 2024-02-09
Move file on backend

### LINKED ISSUE #177: Add a move / copy function

frankenjoe:
Closes #177 

We introduce `backend.Base.move_file()`, which moves a file on the backend to a new location. Similar to #181 a default implementation is provided, which can be overwritten when a backend provides a native way to move files.

### backend.Base

image

### backend.Artifactory

image

### backend.FileSystem

image

### interface.Unversioned

image

### interface.Versioned

image

### Usage

image

...

image

### Development

image

### DIFF
--- a/audbackend/core/backend/artifactory.py
+++ b/audbackend/core/backend/artifactory.py
@@ -281,6 +281,19 @@ class Artifactory(Base):
 
         return paths
 
+    def _move_file(
+            self,
+            src_path: str,
+            dst_path: str,
+            verbose: bool,
+    ):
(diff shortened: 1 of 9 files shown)


hagenw:
Looks good, thanks.

PR #190, opened 2024-03-01
Introduce optional validation step

### LINKED ISSUE #189: Validate download / upload of files
To validate the upload of file to the backend I currently do:

```python
backend.put_file(src_file, dst_file)
if audeer.md5(src_file) != backend.checksum(dst_file):
    # upload failed
    backend.remove_file(dst_file)
    ...
```

Likewise, we can validate the download of a file:

```python
backend.get_file(src_file, dst_file)
if audeer.md5(dst_file) != backend.checksum(src_file):
    # download failed
    os.remove_file(dst_file)
    ...
```

I think it makes sense to move such a check into the functions. Since it comes with an additional cost of calculating the checksum for source and destination file (in case of `put_file()` we already already calculate the checksum of the source file) we could make it an optional step.

In case of a mismatch of the checksum, it makes sense to remove the file. But what should we do next:

1. silently leave the function -> probably not a good idea, since the user assumes the file is there now.
2. notify the user of the failure via return value -> `put_file()` does not have a return value yet, so we could e.g. return a boolean; `get_file()` currently returns the full path to the local file, so here we have to return `None`. To avoid different return types, we could return the full path of the source file in case of `put_file()` and also `None` in case of a failure.
3. Raise an error that informs the user that the operation was not successful.

A fourth option would be to retry the operation first and only if it fails again fall back to one of 1.-3.

frankenjoe:
Closes #189 

Adds `validate` argument to the following functions:

* `copy_file()`
* `get_archive()`
* `get_file()`
* `move_file()`
* `put_archive()`
* `put_file()`

If set to `True` a final check is performed to assert that the new file has the expected checksum. If the check fails, an `InterruptedError` is raised. This comes with additional costs calculating the checksum of the original and new file. In case of `move_file()` the file is copied and only removed if the check was successful.

### DOC

Here are the new docstrings of `backend.Base`:

image

image

image

image

image

image

Similar changes were made to `interface.Unversioned` and `interface.Versioned`

### DIFF
--- a/audbackend/core/backend/base.py
+++ b/audbackend/core/backend/base.py
@@ -41,6 +41,7 @@ class Base:
 
     def _assert_equal_checksum(
             self,
+            *,
             path: str,
             path_is_local: bool,
             path_ref: str,
@@ -192,10 +193,10 @@ class Base:
 
             if validate:
                 self._assert_equal_checksum(
-                    dst_path,
-                    False,
-                    src_path,
-                    False,
+                    path=dst_path,
+                    path_is_local=False,
+                    path_ref=src_path,
+                    path_ref_is_local=False,
                 )
 
     def _create(
@@ -460,10 +461,10 @@ class Base:
 
             if validate:
                 self._assert_equal_checksum(
-                    dst_path,
-                    True,
-                    src_path,
-                    False,
+                    path=dst_path,
+                    path_is_local=True,
+                    path_ref=src_path,
+                    path_ref_is_local=False,
                 )
 
         return dst_path
@@ -876,10 +877,10 @@ class Base:
 
             if validate:
                 self._assert_equal_checksum(
-                    dst_path,
-                    False,
-                    src_path,
-                    True,
+                    path=dst_path,
+                    path_is_local=False,
+                    path_ref=src_path,
+                    path_ref_is_local=True,
                 )
 
     def _remove_file(
--- a/audbackend/core/backend/base.py
+++ b/audbackend/core/backend/base.py
@@ -67,7 +67,7 @@ class Base:
                 location = 'backend'
 
             raise InterruptedError(
-                f"Execution was interrupted because "
+                f"Execution is interrupted because "
                 f"{path} "
                 f"has checksum "
                 f"'{checksum}' "
@@ -151,17 +151,17 @@ class Base:
         have the same checksum.
         If it fails,
         ``dst_path`` is removed and
-        an :class:`InterruptedError` is raised
+        an :class:`InterruptedError` is raised.
 
         Args:
             src_path: source path to file on backend
             dst_path: destination path to file on backend
-            validate: validate file was successfully copied
+            validate: verify file was successfully copied
             verbose: show debug messages
 
         Raises:
             BackendError: if an error is raised on the backend
-            InterruptedError: if validation check fails
+            InterruptedError: if validation fails
             ValueError: if ``src_path`` or ``dst_path``
                 does not start with ``'/'`` or
                 does not match ``'[A-Za-z0-9/._-]+'``
@@ -296,6 +296,7 @@ class Base:
             dst_root: str,
             *,
             tmp_root: str = None,
+            validate: bool = False,
             verbose: bool = False,
     ) -> typing.List[str]:
         r"""Get archive from backend and extract.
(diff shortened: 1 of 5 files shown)


frankenjoe:
Our tests are not working with `pytest==0.8.1` which was released on the weekend.

hagenw:
Looks to me that not our tests, but `doctestplus` is not working with the new `pytest` version:

```
pluggy._manager.PluginValidationError: Plugin 'doctestplus' for hook 'pytest_collect_file'
```

So maybe we wait a few days?

audbackend/core/backend/base.py
```diff
@@ -135,6 +190,14 @@ def copy_file(
                 verbose,
             )
 
+            if validate:
+                self._assert_equal_checksum(
+                    dst_path,
+                    False,
+                    src_path,
+                    False,
+                )
```
hagenw, line 199: The source code is not so easy to read here, as it is not obvious what `False` means.
I guess the easiest fix would be to just write the actual variable names here as well:

```suggestion
                self._assert_equal_checksum(
                    path=dst_path,
                    path_is_local=False,
                    path_ref=src_path,
                    path_ref_is_local=False,
                )
```

If you agree with this, it needs to be changed at the other positions as well.

tests/test_interface_unversioned.py
```diff
@@ -653,3 +655,66 @@ def test_move(tmpdir, src_path, dst_path, interface):
...
+        ) -> str:
+            return ''.join(random.choices(
+                string.ascii_uppercase + string.digits,
+                k=33,
+            ))
+
+    path = audeer.touch(tmpdir, '~.txt')
+    error_msg = 'Execution is interrupted because'
+
+    interface = audbackend.interface.Unversioned(
```
hagenw, line 676: To be on the safe side here, we should maybe add the interface as parameter of the test and than test for all available interfaces?

hagenw (commented):
Looks good, I just added two small suggestions.

Just out of curiosity, can it happen from time to time for some backends that a file gets corrupted that you need this feature?

audbackend/core/backend/base.py
frankenjoe: Ok, I made `_assert_equal_checksum()` a key-word only function.

tests/test_interface_unversioned.py
frankenjoe: Currently, we have two interfaces `Unversioned` and `Versioned` which we test in separate test files. The test you see here is repeated for `Versioned` in `test_interface_versioned.py`. Since the methods expect different arguments, we cannot easily make the interface class a parameter.

hagenw: Ok, then let's stick with the current solution.
Copy and pasting the code to the other test files (at some point we will also have an interface for the legacy functions),
seems also not so nice.

frankenjoe:
> Just out of curiosity, can it happen from time to time for some backends that a file gets corrupted that you need this feature?

Don't know, depends on the backend I guess :)

But giving it a second thought - the main issue are exceptions raised on the backend. When we download a file we are relatively safe because we only move it to the final destination when the download is finished. But with uploads we can end up with incomplete files. However, the strategy we implement here will not apply since the error is raised before the validation check. So maybe we have to consider this? E.g. we could suppress backend errors if `validate=True`. If something goes wrong, the original error is not raised, but we will afterward get a `InterruptedError` and the corrupted file is removed. Maybe we can also catch the original error and print it to give the user a hint what went wrong.

tests/test_interface_unversioned.py
frankenjoe: > at some point we will also have an interface for the legacy functions

Since the legacy methods have the same arguments as the `Versioned` interface, it should be possible to test them mutually.

hagenw:
>  If something goes wrong, the original error is not raised, but we will afterward get a `InterruptedError` and the corrupted file is removed

"something wrong" might also mean that we loose connection to the server, which mean we cannot remove the file. I think in that case the backend needs to support an auto-clean up, e.g. we provide the MD5 sum before starting the upload. When we end (or loose) our connection the backend server than needs to check if the MD5 sum matches and removes the file if not.
As I don't think that most backends support something like this, I don't think this is a possible solution.

So I think, I would not automatically suppress the backend error if `validate=True`. If something goes wrong on the backend, the user will have to act anyway.

frankenjoe:
In principle we could also work with a temporary directory on the backend, which we clean up from time to time. But that really only makes sense if the backend provides a native move function. So also not really a generic solution.

frankenjoe:
> So I think, I would not automatically suppress the backend error if `validate=True`. If something goes wrong on the backend, the user will have to act anyway.

I agree.

The use-case I have in mind is to delete a local copy of file if and only if the file was successfully uploaded to the backend. This can now be achieved by:

```python
try:
    backend.put_file(src_path, dst_path, validate=True)
    os.remove(src_path)   # if we arrive here we can safely remove src_path
except Exception:
    pass
```

Now this is even safe if the backend does not raise an error in case something goes wrong during the upload.

hagenw:
OK, looks like a valid example to me.
This means we are ready here and don't need further changes to the code, correct?

frankenjoe:
I would say yes.

PR #192, opened 2024-03-04
TST: depend on pytest<8.1.0

frankenjoe:
`doctestplus` is not compatible with `pytest==8.1.0`, see https://github.com/audeering/audbackend/pull/190#issuecomment-1976241914

### DIFF
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -1,3 +1,3 @@
-pytest
+pytest<8.1.0  #  INTERNALERROR> Argument(s) {'path'} are declared in the hookimpl but can not be found in the hookspec
 pytest-cov
 pytest-doctestplus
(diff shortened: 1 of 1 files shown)


hagenw:
I reported it upstream at https://github.com/scientific-python/pytest-doctestplus/issues/239.

As we are working in the development branch at the moment, I agree with just pinning the `pytest` version until `doctestplus` is fixed.

hagenw:
As expected, this caused some uproar in the community.

In the meantime, `pytest` v8.1.0 was yanked (https://github.com/pytest-dev/pytest/issues/12069) and we got a new release of `doctestplus` (https://github.com/scientific-python/pytest-doctestplus/issues/239#issuecomment-1976973156). I would propose we wait until they repeat the actual `pytest` v8.1.0 release and do a new pull request then removing the upper limit for `pytest`.

PR #193, opened 2024-03-21
Use ruff formatter as pre-commit

hagenw:
As `ruff` now supports formatting of code like `black`, we extend `pre-commit` to also format the code automatically using `ruff`.

The biggest changes with the autoformatted code are:
* we now use `"` instead of `'`
* the default line length is now 88 chars
* we use less indention for multi-line function arguments, if the function has a docstring


frankenjoe:
> * we now use `"` instead of `'`

I personally prefer `'` as it makes the code easier to read. But I suppose this is now standard with black, so we don't want to mess around with it.

PR #194, opened 2024-03-22
Add audbackend.interface.Maven

### LINKED ISSUE #168: Add VersionedLegacy interface
In https://github.com/audeering/audbackend/pull/167 we introduced the concept of interfaces as an abstraction layer between backends and the user. This provides the possibility to also handle the legacy versioning (as we use in `audb`) as an extra interface.
The filestructure on the backend is:
```
<root>/<base>/<version>/<base>-<version><ext>
```

Instead of `VersionedLegacy` we might also pick another name for it (e.g. maybe there is a name for that structure) and then we have it as a feature and not only legacy code.

The only problem I see is that you might get in troubles when using the `Versioned` interface on a repo that was created using `VersionedLegacy` (I tried this ones for an `audb` repo. The problem was that I had a hard time to understand the problem after seeing some strange error method when executing a method. Also not all methods were failing). So it might be that we also have to consider https://github.com/audeering/audbackend/issues/166, even though I agree that it is also a cool feature to be able to store versioned and unversioned files on the same repo as you can do with the `Versioned` and `Unversioned` interfaces.

hagenw:
Closes #168 

This adds an extra interface to handle our legacy versioned file structure.

I'm still not completely convinced by the `Maven` name. If you have a better proposal, I'm open to change the name.

---

The API documentation is currently slightly broken as it uses the name `versioned` to refer to the interface, but the problem exists also for the `Versioned` and `Unversioned` API, which also uses a mix of `interface` and their actual variable names. I propose to fix this in another pull request, after this one is finished.

Excerpt from the API documentation:

image

---

For now I haven't added a new section to the usage documentation, but updated the "Legacy backends" page to

image

### DIFF
--- a/tests/test_interface_maven.py
+++ b/tests/test_interface_maven.py
@@ -1,10 +1,71 @@
 import os
+import re
+import stat
 
 import pytest
 
 import audeer
 
 import audbackend
+
+
+@pytest.mark.parametrize(
+    "interface",
+    [("file-system", audbackend.interface.Maven)],
+    indirect=True,
+)
+def test_errors(tmpdir, interface):
+    # Ensure we have one file and one archive published on the backend
+    archive = "/archive.zip"
+    local_file = "file.txt"
+    local_path = audeer.touch(audeer.path(tmpdir, local_file))
+    remote_file = f"/{local_file}"
+    version = "1.0.0"
+    interface.put_file(local_path, remote_file, version)
+    interface.put_archive(tmpdir, archive, version, files=[local_file])
+
+    # Create local read-only file and folder
+    file_read_only = audeer.touch(audeer.path(tmpdir, "read-only-file.txt"))
+    os.chmod(file_read_only, stat.S_IRUSR)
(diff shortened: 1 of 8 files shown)


tests/test_interface_maven.py
```diff
@@ -0,0 +1,155 @@
...
+        ("/.sub/", False, None, hidden),
+        ("/.sub/", True, None, hidden_latest),
+        ("/file.bar", False, None, root_bar),
+        ("/file.bar", True, None, root_bar_latest),
+        ("/sub/file.foo", False, None, sub),
+        ("/sub/file.foo", True, None, sub_latest),
+        ("/sub/file.foo", False, "file.*", sub),
+        ("/sub/file.foo", True, "file.*", sub_latest),
+        ("/sub/file.foo", False, "*.bar", []),
+        ("/sub/file.foo", True, "*.bar", []),
```
hagenw, line 147: Those four cases are broken for `audbackend.interface.Versioned` (https://github.com/audeering/audbackend/issues/196), but work here.

Issue #199, opened 2024-03-25
Add backend method open() and close()


hagenw:
Copied from https://github.com/audeering/audbackend/issues/197#issuecomment-2017517185

[...] it is very common to open and close connections to a storage drive. So maybe we should actually introduce both functions, e.g.:

```python
backend = audbackend.backend.MyBackend(host, repo)
backend.open()
# do something with the backend
backend.close()
```

or alternatively:

```python
with audbackend.backend.MyBackend(host, repo) as backend:
     # do something with the backend
```

hagenw:
@frankenjoe I will continue working on #197, so I assign this to you (see also https://github.com/audeering/audbackend/pull/200#issuecomment-2024981522)

hagenw:
Solved by #204

Issue #197, opened 2024-03-22
Simplify accessing a repository


hagenw:
After the introduction of the interfaces, we removed the return value of `audbackend.create()` (https://github.com/audeering/audbackend/pull/174) after [this discussion](https://github.com/audeering/audbackend/pull/171#issuecomment-1909725543).
I think this was a good idea, but I would even go further and propose that `audbackend.access()` is no longer needed.

To access a backend, you can do at the moment:
```python
interface = audbackend.access(backend, host, repo)
```
It will return the default versioned interface, but a user might not be aware of this.

To access a different interface, that might even have arguments:
```python
interface = audbackend.access(
    backend,
    host,
    repo,
    interface=audbackend.core.FancyInterface,
    interface_kwargs={"myarg": 0},
)
```
This feels a little bit cumbersome and also strange as it returns an interface, and you have to provide an interface as argument.

I would propose instead to remove (deprecate) `audbackend.access()` and change how an interface is initialized to:
```python
interface = audbackend.core.FancyInterface(backend, host, repo, myarg=0)
```
Then it will be always obvious which interface will be returned, and a user needs only to look up the documentation of the interface, not of the interface and `audbackend.access()` together.

frankenjoe:
I stopped using `audbackend.access()` already some time ago, instead in my projects I simply do:

``` python
backend = audbackend.MyBackend(host, repo)
interface = audbackend.interface.MyInterface(backend)
```

But maybe we should think about also deprecating `create()`, `delete()` and `register()` then and instead introduce `audbackend.Base.create()` and `audbackend.Base.delete()`?

frankenjoe:
Otherwise, if we decide to keep the API functions. Then I would propose we deprecate `audbackend.access()` and instead introduce a new function like `audbackend.backend()` that works like `access()` but returns `audbackend.backend.Base`. Then a user would e.g. do:

```python
backend = audbackend.backend(name, host, repo)
interface = audbackend.interface.Versioned(backend)
```

frankenjoe:
btw: if we deprecate `access()` we can immediately deprecate `interface.Unversioned`, too :)

hagenw:
One advantage (or maybe the motivation) of `create()`, `delete()`, and `register()` are that a user does not need to deal with the backend classes. I had this in mind when I proposed to use `audbackend.interface.MyInterface(name, host, repo)` to replace `access()` as it would be in line with that approach.

[...]

PR #200, opened 2024-03-26
Add Backend class methods create() and delete()

hagenw:
Relates to #197 

I think it makes sense to split the tasks from #197 into more than one pull request.
Here, I start with adding the class methods `audbackend.backend.Base.create()` and `audbackend.backend.Base.delete()`,
stop using `audbackend.access()` to get an interface to a backend, and no longer register backends.

I also have ~~removed the `_access()` method and~~ updated the "Usage", "Developer guide", and "Legacy backends" documentation.
In the "Developer guide" I have commented out large parts of the `SQLite` example, as it needs #199 in order to work without error. And I would propose to update it when adding those functionality.

Example docstrings from `audbackend.backend.FileSystem`:

image

image

### DIFF
--- a/audbackend/core/backend/base.py
+++ b/audbackend/core/backend/base.py
@@ -9,6 +9,11 @@ from audbackend.core import utils
 from audbackend.core.errors import BackendError
 
 
+# Backend type to annotate Backend.create().
+# From Python 3.11 on we can use typing.Self instead.
+# See https://stackoverflow.com/a/71118499
+Backend = typing.TypeVar("Backend", bound="Base")
+
 class Base:
     r"""Backend base class.
 
@@ -211,7 +216,7 @@ class Base:
         cls,
         host: str,
         repository: str,
-    ) -> "Backend":
+    ) -> Backend:
         r"""Create repository.
 
         Creates ``repository``
--- a/audbackend/core/backend/base.py
+++ b/audbackend/core/backend/base.py
@@ -3,16 +3,16 @@ import os
 import tempfile
 import typing
 
+# Backend type to annotate Backend.create().
+# From Python 3.11 on we can use typing.Self instead.
+# See https://stackoverflow.com/a/71118499
+from typing_extensions import Self
+
 import audeer
 
 from audbackend.core import utils
 from audbackend.core.errors import BackendError
 
-
-# Backend type to annotate Backend.create().
-# From Python 3.11 on we can use typing.Self instead.
-# See https://stackoverflow.com/a/71118499
-Backend = typing.TypeVar("Backend", bound="Base")
 
 class Base:
     r"""Backend base class.
@@ -216,7 +216,7 @@ class Base:
         cls,
         host: str,
         repository: str,
-    ) -> Backend:
+    ) -> Self:
         r"""Create repository.
 
         Creates ``repository``
--- a/audbackend/core/backend/base.py
+++ b/audbackend/core/backend/base.py
@@ -206,6 +206,32 @@ class Base:
         """
         raise NotImplementedError()
 
+    @classmethod
+    def create(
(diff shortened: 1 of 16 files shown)


hagenw:
OK, makes sense, I added `_access()` again.

audbackend/core/backend/base.py
```diff
@@ -3,6 +3,11 @@
 import tempfile
 import typing
 
+# Backend type to annotate Backend.create().
+# From Python 3.11 on we can use typing.Self instead.
+# See https://stackoverflow.com/a/71118499
+from typing_extensions import Self
```
frankenjoe, line 9: Since locally I am running Python 3.10 it gives me:

```
ModuleNotFoundError: No module named 'typing_extensions'
```

hagenw: I had to add it as a dependency, but it should be small.

audbackend/core/backend/base.py
```diff
@@ -206,6 +211,32 @@ def _create(
...
+
+        Raises:
+            BackendError: if an error is raised on the backend,
+                e.g. repository exists already
+                or cannot be created
+
+        """
+        backend = cls(host, repository)
+        utils.call_function_on_backend(backend._create)
+        return backend
```
frankenjoe, line 238: I think we have recently changed `audbackend.create()` to not (officially) return the `backend` object and only kept the return argument for backward compatibility. The argument was that in your code you should not do:

```python
backend = MyBackend.create()
# now do something with backend
```

Because running the code again will fail (since the repository now exists).

If we decide to switch back to returning the backend, we should document it in the `Returns` section of the docstring.

tests/test_interface_maven.py
```diff
@@ -7,11 +7,18 @@
 import audeer
 
 import audbackend
+from audbackend.backend import FileSystem
```
frankenjoe, line 10: I guess the motivation for this import is to make the code shorter, right? Maybe we should then also apply it in the other test files, e.g. in `tests/test_backend_filesystem.py` we are still using `audbackend.backend.FileSystem`.

frankenjoe:
We are still calling `audbackend.create()` and `audbackend.delete()` in `audbackend/core/conftest.py`. Maybe change that, too?

hagenw:
> We are still calling `audbackend.create()` and `audbackend.delete()` in `audbackend/core/conftest.py`. Maybe change that, too?

if we change this `audb.available()` will no longer show any output in the docstring. So I would prefer to change this when we deprecate `audb.available()` and remove it from the API documentation (or at least remove the example from the documentation).

audbackend/core/backend/base.py
hagenw: Good point, I changed it to not return the backend (and removed `typing_extensions` from the package dependencies).
Until #199 is implemented, our `SQLite` example in the developer docs does then no longer work (don't know why it worked before with the returned backend ;) ). As stated in the description of this pull request, I uncommented and propose to undo those changes and update the example when we solve #199.

tests/test_interface_maven.py
hagenw: I reverted this and use now everywhere `audbackend.backend.FileSystem`, ...

audbackend/core/backend/base.py
hagenw: It worked before, because we have `self._db = sl.connect(self._path)` inside `_create()`, which returned then a backend with an already connected database. But this also highlights, that it was not a good idea by me to add a return value to `create()` as accessing and leaving a backend should indeed be handled by #199.

frankenjoe:
Looks good! Will you also take care of #199?

frankenjoe:
Otherwise I can also give it a try :)

PR #201, opened 2024-03-27
CI: update actions to switch to Node.js 20

hagenw:
As discussed at https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/, we need to update all Github Actions that are still using Node.js 16.

This can also be seen when inspecting a [random Actions page](https://github.com/audeering/audbackend/actions/runs/8448176848?pr=200):

image

We fix this here by updating to:

* actions/checkout@v4
* actions/setup-python@v5
* codecov/codecov-action@v4
* softprops/action-gh-release@v2

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -19,7 +19,7 @@ jobs:
     - uses: actions/checkout@v4
 
     - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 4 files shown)


hagenw:
Deprecation warnings are gone with this update, compare https://github.com/audeering/audbackend/actions/runs/8448387732

PR #202, opened 2024-03-28
Remove audbackend.available()

hagenw:
Relates to #197 

This pull request proposes to remove `audbackend.available()` as we cannot maintain its current behavior/results with the new approach of accessing backends by instantiating them, introduced in #200.

There might be ways to preserve its original behavior, but I don't think this is worse the effort.

I also updated `audbackend/core/conftest.py` as proposed in https://github.com/audeering/audbackend/pull/200#issuecomment-2024615148.

### DIFF
--- a/audbackend/__init__.py
+++ b/audbackend/__init__.py
@@ -1,7 +1,6 @@
 from audbackend import backend
 from audbackend import interface
 from audbackend.core.api import access
-from audbackend.core.api import available
 from audbackend.core.api import create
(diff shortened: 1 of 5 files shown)


frankenjoe:
@hagenw ups, just realized you had not yet assigned the PR to me. Sorry, I hope you were finished.

hagenw:
Yes, I was just waiting for the CI pipelines to finish. All fine.

PR #203, opened 2024-03-28
Deprecate audbackend.register()

hagenw:
Relates to #197 

This deprecates `audbackend.register()` and marks it to be removed in version 2.2.0, assuming that the current `dev` branch will be released as version 2.0.0.

image

### DIFF
--- a/audbackend/core/conftest.py
+++ b/audbackend/core/conftest.py
@@ -44,10 +44,16 @@ def prepare_docstring_tests(doctest_namespace):
         current_dir = os.getcwd()
         os.chdir(tmp)
 
+        warning = (
+            "register is deprecated and will be removed with version 2.2.0. "
+            "Use backend classes directly instead."
+        )
+
         file = "src.pth"
         audeer.touch(file)
 
-        audbackend.register("file-system", DoctestFileSystem)
+        with pytest.warns(UserWarning, match=warning):
+            audbackend.register("file-system", DoctestFileSystem)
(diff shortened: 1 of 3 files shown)


frankenjoe:
Since you started removing docstring examples from other API functions, wouldn't it make sense to extend this PR to deprecate also the other functions?

hagenw:
I thought it would be better to have single commits for deprecating the other functions as well, but if you prefer, we can also integrate it here.

frankenjoe:
I think we can go with one PR since those functions are closely linked.

hagenw:
OK, it also makes it easier for me to combine all into one pull request.
Should be ready again for review.

audbackend/core/conftest.py
```diff
@@ -83,9 +67,8 @@ def prepare_docstring_tests(doctest_namespace):
 
         yield
 
-        audbackend.delete("file-system", "host", "repo")
-        audbackend.delete("file-system", "host", "repo-unversioned")
-        audbackend.register("file-system", audbackend.backend.FileSystem)
+        DoctestFileSystem.delete("host", "repo")
+        DoctestFileSystem.delete("host", "repo-unversioned")
         doctest_namespace["create"] = audbackend.create
```
frankenjoe, line 72: I think this line is no longer needed.

hagenw: I removed the line in a64b4ce

PR #204, opened 2024-03-28
Add backend methods open() and close()

### LINKED ISSUE #199: Add backend method open() and close()

frankenjoe:
Closes #199


docs/usage.rst
```diff
@@ -47,16 +47,36 @@ We select the :class:`audbackend.backend.FileSystem` backend.
...
+:meth:`audbackend.backend.open`
+and
+:meth:`audbackend.backend.close`
+it is good practise to use a with statement
+(not applicable in a notebook).
+
+.. jupyter-execute::
+
+    backend = audbackend.backend.FileSystem("./host", "repo")
+    backend.open()
```
hagenw, line 67: This seems a little bit strange to me. I see that there are backends that need to be opened and closed, but why should I have to open a file system backend. On those simple backends I would have expected that a simple `backend = audbackend.backend.FileSystem("./host", "repo")` should be sufficient.

frankenjoe: It is, except that it checks if the folder exists it does nothing. But I would still show it here since with other backends (e.g. the SQL one) it is important, so I think it is good practice to always call `open()` and `close()` (or if possible use a `with` statement).

docs/usage.rst
```diff
@@ -172,21 +192,19 @@ We can remove files.
...
+we close the connection to the backend.
 
 .. jupyter-execute::
 
-    audbackend.backend.FileSystem.delete("host", "repo")
-
+    backend.close()
 
-We can check if a repository exists
-by inspecting its root path.
```
hagenw, line 184: As we no longer can call exists, I think we also need a method/way to check if the repository exists.
At the moment `open()` seems to be the solution? If yes, we should maybe add it here instead of deleting the example.
On the other hand, `open()` feels not very intuitive to me to check if the repo exists.

frankenjoe: Yes, maybe such a function makes sense. It's a bit unfortunate that we called the function to check if a file exists `exists()` and not `exists_file()` as with all the other functions like `put_file()`, etc.

frankenjoe: Btw: it's a bit strange that it was working before. When I delete a repository I would expect that the folder is removed, too. But it seems that this is not the case since `exists('/')` was not raising an error when called after `delete()`.

frankenjoe: And apart from that it is also strange that is is possible to call `exists()` with `/` which is not the path to a file.

hagenw: `exists('/')` returned `False` for file-system as the folder did not exists. What is not deleted by `delete()` is the host (as for some backends we can not delete the host), whereas `create()` also cretaes the host folder for `FileSystem`.

docs/usage.rst
hagenw: Having in mind that it is much more fun to work with `auglib` since you don't have to use it with `with`, I would highly appreciate if `open()`, `close()` / `with` are only used for backends that require it. For all others it should still be there (to not raise an error when using the same code as for another backend), but I would assume that a user needs nothing to implement if a backend does not need `open()`, `close()`. The later is already implemented in this pull request as we have `open()`, `close()` in `audbackend.backend.Base` that do nothing. In `audbackend.backend.FileSystem` we have implemented `_open()` at the moment, which I think we should remove. Again this then boils down to the problem at which stage and in which method we should check if a repo actually exists. E.g. if we integrate authentication into `open()` we can only do it after the connection is already established.

frankenjoe: Yes, implementing `_open()` and `_close()` is optional. But for some backends it can be important to call it. So I think it's good practice to always call it. I think it also makes sense to use the `open()` function to check if a repository exists or - in case of Artifactory - if a connection can be established.

docs/usage.rst
```diff
@@ -47,16 +47,36 @@ We select the :class:`audbackend.backend.FileSystem` backend.
...
+Later on,
+we will use
+:meth:`audbackend.backend.close`
+to close the connection.
+Instead of explicitly calling
+:meth:`audbackend.backend.open`
+and
+:meth:`audbackend.backend.close`
+it is good practise to use a with statement
+(not applicable in a notebook).
```
hagenw, line 62:
```suggestion
we can access it by instantiating the backend class.
Some backends also need to establish a connection.
This can be achieved using a ``with`` statement,
or by calling ``backend.open()`` at the beginning,
and ``backend.close()`` at the end.
If you are unsure
weither your backend needs to establish a connection,
just always do it.
```

docs/usage.rst
```diff
@@ -47,16 +47,36 @@ We select the :class:`audbackend.backend.FileSystem` backend.
...
+.. jupyter-execute::
+
+    backend = audbackend.backend.FileSystem("./host", "repo")
+    backend.open()
+
+After establishing a connection
+we could directly execute read and write operations
+on the backend object.
+However,
+here we use :class:`audbackend.interface.Unversioned`
```
hagenw, line 73:
```suggestion
However,
we recommend to always use :mod:`interfaces <audbackend.interface>`
to communicate with a backend.
Here, we use :class:`audbackend.interface.Unversioned`
```

docs/usage.rst
```diff
@@ -172,21 +192,19 @@ We can remove files.
...
-We can check if a repository exists
-by inspecting its root path.
+And delete the whole repository
+with all its content.
 
 .. jupyter-execute::
 
-    backend = audbackend.backend.FileSystem("host", "repo")
-    backend.exists("/")
+    audbackend.backend.FileSystem.delete("host", "repo")
```
hagenw, line 208:
```suggestion

For :class:`audbackend.backend.FileSystem`
we can test
if the repository exists
by trying to connect to it. 

.. jupyter-execute::

    try:
        backend.open()
    except audbackend.BackendError as ex:
        display(str(ex.exception))

```

hagenw (commented):
The introduction to the `SQLite` backend currently reads like this:

image

Would be nice to discuss there already, that we also need to add the `_open()` and `_close()` methods for this backend.

docs/usage.rst
hagenw: OK, I'm fine with staying with `backend.open()` here, but suggested a few changes to the text below.

docs/usage.rst
hagenw: Yes and no. You are right that we state that `path` should be a file path, but we do also not discuss what should happen if you provide `/` as path in the docstring:

image

hagenw: I think we still have two options here (that we should handle in a separate pull request):
* raise an error for `"/"` as `path` and extend the `Raises` section accordingly
* check for `"/"` if the repository exists and extend the docstring accordingly

frankenjoe: > raise an error for "/" as path and extend the Raises section accordingly

Not specifically `"/"`, but any path that ends on `"/"`. One solution would be to extend `utils.check_path()` accordingly and then extend the description of the first `ValueError`. This applies to all functions except `ls()`, which is the only function that accepts a sub-path.

> check for "/" if the repository exists and extend the docstring accordingly

I thought about that, too. But what I do not like about this solution is that usually you call `open()` first (at least on some backends) before you can call `exists()`. But then with `'/'` as argument you would call it **before** `open()` to check if you can call `open()`. That is slightly confusing.

docs/usage.rst
frankenjoe: Should we mention that it behaves exactly like calling the methods directly on `backend`?

hagenw: No, I think this doesn't matter as a user should anyway not directly interact with the backend object.

frankenjoe:
> Would be nice to discuss there already, that we also need to add the `_open()` and `_close()` methods for this backend.

Good catch, updated to;

Screenshot from 2024-04-02 14-58-47

docs/usage.rst
frankenjoe: Screenshot from 2024-04-02 15-01-14

PR #214, opened 2024-04-08
Deprecate audbackend.access()

### LINKED ISSUE #211: Deprecate audbackend.access()
Relates to https://github.com/audeering/audbackend/issues/197

In #203 we have not deprecated `audbackend.access()` yet. But I guess we should do so, too.

hagenw:
Closes #211 

This deprecates `audbackend.access()`.

image

### DIFF
--- a/audbackend/core/api.py
+++ b/audbackend/core/api.py
@@ -73,7 +73,13 @@ def access(
         and will be removed in version 2.2.0.
         Repositories on backends are instead accessed
         by instantiating the corresponding backend class,
-        e.g. :meth:`audbackend.backend.FileSystem(host, repo)`.
+        and connecting to it using the ``open()`` method,
+        e.g.
+
+        .. code-block:: python
+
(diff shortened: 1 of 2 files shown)


audbackend/core/api.py
```diff
@@ -63,6 +67,14 @@ def access(
     on the backend with alias ``name``
     (see :func:`audbackend.register`).
 
+    .. Warning::
+
+        ``audbackend.access()`` is deprecated
+        and will be removed in version 2.2.0.
+        Repositories on backends are instead accessed
+        by instantiating the corresponding backend class,
+        e.g. :meth:`audbackend.backend.FileSystem(host, repo)`.
```
frankenjoe, line 76: Doesn't look right to me:

image

One could also argue that the replacement for `access` is `open`.

hagenw: I updated the warning message to

image

Issue #162, opened 2023-12-21
Add auth argument to provide username, password to backend


hagenw:
The only way to authenticate at a backend at the moment, is to provide the credentials in a config file or an environment variable. This is fine in most cases, but not in all, e.g. when accessing the same backend with different users, one needs to set and delete environment variables.

This could be avoided by adding an `auth` variable (as in `requests.get()`) to all functions/methods that have to authenticate:
* `Backend.__init__()`
* `access()`
* `create()`
* `delete()`

The `auth` argument would then expect the tuple `(username, password)`.

frankenjoe:
Implemented in #215

PR #215, opened 2024-04-10
Better authentication handling in backends

### LINKED ISSUE #161: Add public authentication() method to Artifactory backend
In order to deprecate the [audfactory](https://github.com/audeering/audfactory) package we should add the private function `audbackend.core.artifactory._authentication()` as a public class method.

```python
username, password = audbackend.Artifactory.authentication(host)
```

This would a user allow to test which username and password would be used to authenticate with the host,
useful for debugging authentication issues, see https://audeering.github.io/audfactory/usage.html#authentication.
And it would allow to get the username, password needed for rest API requests, as in [audfactory.rest_api_get()](https://github.com/audeering/audfactory/blob/6fc6afe2e0ab53c25f7ebd946c584f5756c0466a/audfactory/core/api.py#L353-L354):
```python
    username, apikey = authentification(url)
    return requests.get(url, auth=(username, apikey))
```

hagenw:
Closes #161, #162

---

### * Add class method `audbackend.backend.Artifactory.get_authentication()`

This tackles #161 by providing a public class method to read authentication tokens from a config file or environment variables.

image

---

### * Add `authentication` argument and attribute to `audbackend.backend.Base`

This tackles #162 by adding an `authentication` argument to `__init__()`, `create()`, and `delete()` of `audbackend.backend.Base`.

In `audbackend.backend.FileSystem` I then do not add `authentication` as an argument to `__init__()`. Don't know if it would be better to always have it, but from a user perspective I found it easier this way.

---

### * Speedup `audbackend.backend.Artifactory`

`audbackend.backend.Artifactory` now only applies the authentication when calling `open()`. Before it did the authentication every time a path was requested on the backend, e.g. every time `backend.exists(path)` was called. Now, we store the path to the repo in `self._repo`, when calling `open()`, and can create all other paths based on top of that.

When asking if 4 files exists on the backend this reduced the execution time for me from **0.872 s** to **0.178 s**.

I also updated the code of the `interface` test fixture in `tests/conftest.py` to directly delete the repository on an Artifactory backend at the end of the test without the need to first re-authenticate.

<details><summary>Benchmark code</summary>

```python
import time

import audbackend


backend = audbackend.backend.Artifactory("https://audeering.jfrog.io/artifactory", "data-public")
backend.open()
interface = audbackend.interface.Maven(backend)

t0 = time.time()
interface.exists("/emodb", "1.4.1")
interface.exists("/emodb/db", "1.4.1")
interface.exists("/emodb/meta", "1.4.1")
interface.exists("/emodb/media", "1.4.1")
t = time.time()
print(f"{t - t0:.3f} s")
```

</details>

image

### * Add `audbackend.backend.Artifactory.path()`

Before we had `audbackend.backend.Artifactory._path()`, which was returning an `artifactory.ArtifactoryPath` object that can be used to interact with the path on the backend. As this might also be helpful for a user (compare https://github.com/audeering/audb/pull/386), I decided to change it into a public method.

image


hagenw:
I converted back to draft, as I would first like to solve one remaining problems with the tests.

hagenw:
I was able to fix the test related problems, and the pull request is again ready for review.

hagenw:
We could also speed up the tests on the Artifactory backend even further by just open one connection to the server and reuse it, but I would propose to not handle this in this pull request and created #218 to track it.

frankenjoe:
> `audbackend.backend.Artifactory` now only applies the authentication when calling `open()`. Before it did the authentication every time a path was requested on the backend, e.g. every time `backend.exists(path)` was called. Now, we store the path to the repo in `self._repo`, when calling `open()`, and can create all other paths based on top of that.
> 
> When asking if 4 files exists on the backend this reduced the execution time for me from **0.872 s** to **0.178 s**.

This is a great improvement and should lead to a significant speed up in `audb` when publishing / loading databases with many small files.

hagenw:
Yes, this would also be my hope, maybe I should try to benchmark that already.

frankenjoe:
> In `audbackend.backend.FileSystem` I then do not add `auth` as an argument to `__init__()`. Don't know if it would be better to always have it, but from a user perspective I found it easier this way.

Seems ok to me, to only add it to classes where it is actually needed.

audbackend/core/backend/artifactory.py
```diff
@@ -109,45 +69,106 @@ def _download(
...
     """  # noqa: E501
 
     def __init__(
         self,
-        host,
-        repository,
+        host: str,
+        repository: str,
+        *,
+        auth: typing.Tuple = None,
```
frankenjoe, line 86:
```suggestion
        auth: typing.Tuple[str, str] = None,
```

audbackend/core/backend/base.py
```diff
@@ -25,11 +26,15 @@ def __init__(
         self,
         host: str,
         repository: str,
+        *,
+        auth: typing.Any = None,
```
frankenjoe, line 30: Since we also say "repository" and not "repo", I would use the full word `authentication` here, too.

tests/bad_file_system.py
```diff
@@ -1,9 +1,29 @@
+import audeer
+
 import audbackend
 
 
 class BadFileSystem(audbackend.backend.FileSystem):
     r"""Imitates a corrupted file system."""
 
+    def put_file(
```
frankenjoe, line 9: Why do we need to override `Base.put_file()` here?

tests/test_backend_artifactory.py
```diff
@@ -254,9 +204,9 @@ def test_maven_file_structure(
     src_path = audeer.touch(audeer.path(tmpdir, "tmp"))
     interface.put_file(src_path, file, version)
 
-    url = f"{str(interface.backend._repo.path)}{expected}"
-    url_expected = interface.backend._expand(
-        interface._path_with_version(file, version),
+    url = f"{str(interface.backend.path(expected))}"
```
frankenjoe, line 207:
```suggestion
    url = str(interface.backend.path(expected))
```

audbackend/core/backend/artifactory.py
```diff
@@ -109,45 +69,106 @@ def _download(
...
+        Note, API keys are deprecated
+        and will no longer work
+        with newer versions of Artifactory.
+
+        To get the username,
+        password/access token combination,
+        the function looks first
+        for the two environment variables
+        ``ARTIFACTORY_USERNAME`` and
+        ``ARTIFACTORY_API_KEY``.
```
frankenjoe, line 115: Do we also use the variable `ARTIFACTORY_API_KEY` for access tokens?

audbackend/core/backend/base.py
```diff
@@ -261,14 +268,21 @@ def create(
         Args:
             host: host address
             repository: repository name
+            auth: access token
+                for possible authentication,
+                e.g. username, password tuple
```
frankenjoe, line 273: "access token" already sounds quite specific to me. Maybe we keep it more general: 

```suggestion
            auth: object used for authentication,
                e.g. a tuple with username and password
```

audbackend/core/backend/artifactory.py
hagenw: Yes, it's all very messy with names for authentication in Artifactory.
They will remove support for API keys and have it replaced with access tokens, which are very similar but automatically expire after some time.

Do not make it too complicated I decided for now to not add another environment variable. But I agree that it is unfortunate that the name will not match (API key != access token). Would you propose to add `ARTIFACTORY_ACCESS_TOKEN` and then simply first look for that, and if empty/missing look for `ARTIFACTORY_API_KEY` next?

audbackend/core/backend/base.py
hagenw: Yes, good idea. I changed the `auth` argumet and attribute to `authentication`, and renamed the `authentication()` class method to `get_authentication()`.

audbackend/core/backend/artifactory.py
hagenw: No, I'm also not convinced by their new approach. I as an admin have to create new access tokens and update them here on Github.

At the moment our tests here are failing as I still have a bug in the code, which I introduced when renaming `auth` to `authenticaton`. I'm working to solve this at the moment.

---

In this pull request, I would then stay with `ARTIFACTORY_API_KEY`, but created https://github.com/audeering/audbackend/issues/219 to solve this in the future.

hagenw: I found the error, and have fixed it.

PR #217, opened 2024-04-16
Make open a backend mandatory

### LINKED ISSUE #216: What should we do if a backend requires open()
In https://github.com/audeering/audbackend/pull/204 we introduced `open()` and `close()` methods.

At the moment, I'm cleaning up the authentication process for Artifactory and thought that it would make sense to move the authentication from `__init__()` to `open()`. The problem, I'm experiencing with this approach is that I don't know how I should handle the case if a user does not call `open()`, but tries to directly access some file, e.g. with `exists()`. Should I raise a `audbackend.BackenError` and state that the backend needs to be first opened?

hagenw:
Closes #216 

This adds a new `opened` attribute to the backends to track if the backend is currently open or not:

image

In addition, every backend method that needs access to the actual backend (e.g. `sep()` does not need this), now raises a `RuntimeError` if the backend is not opened.

Example "Raises" section for the new `RuntimeError` that is raised when the backend was not opened:

image

image

### DIFF
--- a/audbackend/core/backend/artifactory.py
+++ b/audbackend/core/backend/artifactory.py
@@ -208,9 +208,8 @@ class Artifactory(Base):
         self,
     ):
         r"""Delete repository and all its content."""
-        self.open()
-        self._repo.delete()
-        self.close()
+        with self:
+            self._repo.delete()
 
     def _exists(
         self,
--- a/audbackend/core/backend/artifactory.py
+++ b/audbackend/core/backend/artifactory.py
@@ -142,14 +142,6 @@ class Artifactory(Base):
     ):
         super().__init__(host, repository)
 
-        self._username, self._api_key = _authentication(host)
-        path = _artifactory_path(
-            self.host,
-            self._username,
-            self._api_key,
-        )
-        self._repo = path.find_repository_local(self.repository)
-
     def _checksum(
         self,
         path: str,
@@ -191,20 +183,16 @@ class Artifactory(Base):
         self,
     ):
         r"""Access existing repository."""
-        if self._repo is not None:
-            utils.raise_file_exists_error(str(self._repo.path))
-
-        path = _artifactory_path(
-            self.host,
-            self._username,
-            self._api_key,
-        )
-        self._repo = dohq_artifactory.RepositoryLocal(
+        username, api_key = _authentication(self.host)
+        path = _artifactory_path(self.host, username, api_key)
+        repo = dohq_artifactory.RepositoryLocal(
             path,
             self.repository,
             package_type=dohq_artifactory.RepositoryLocal.GENERIC,
         )
-        self._repo.create()
+        if repo.path.exists():
+            utils.raise_file_exists_error(str(repo.path))
+        repo.create()
 
     def _date(
         self,
@@ -220,7 +208,9 @@ class Artifactory(Base):
         self,
     ):
         r"""Delete repository and all its content."""
+        self.open()
         self._repo.delete()
+        self.close()
 
     def _exists(
         self,
@@ -288,8 +278,15 @@ class Artifactory(Base):
         self,
     ):
         r"""Open connection to backend."""
+        self._username, self._api_key = _authentication(self.host)
+        path = _artifactory_path(
+            self.host,
+            self._username,
+            self._api_key,
+        )
+        self._repo = path.find_repository_local(self.repository)
         if self._repo is None:
-            utils.raise_file_not_found_error(str(self._repo.path))
+            utils.raise_file_not_found_error(self.repository)
 
     def _owner(
         self,
--- a/audbackend/core/backend/base.py
+++ b/audbackend/core/backend/base.py
@@ -31,7 +31,7 @@ class Base:
         self.repository = repository
         r"""Repository name."""
         self.opened = False
-        r"""If backend is opened."""
+        r"""If a connection to the repository has been established."""
 
     def __enter__(self):
         r"""Open connection via context manager."""
--- a/audbackend/core/backend/base.py
+++ b/audbackend/core/backend/base.py
@@ -9,6 +9,9 @@ from audbackend.core import utils
 from audbackend.core.errors import BackendError
 
 
+backend_not_opened_error = "'Backend.open()' needs to be run first."
+
+
 class Base:
     r"""Backend base class.
 
@@ -25,6 +28,8 @@ class Base:
         r"""Host path."""
         self.repository = repository
         r"""Repository name."""
+        self.opened = False
+        r"""If backend is opened."""
 
     def __enter__(self):
         r"""Open connection via context manager."""
@@ -103,13 +108,16 @@ class Base:
             MD5 checksum
 
         Raises:
-            BackendError: if an error is raised on the backend,
+            BackendError: if backend was not opened,
+                or an error is raised on the backend,
                 e.g. ``path`` does not exist
             ValueError: if ``path`` does not start with ``'/'``,
                 ends on ``'/'``,
                 or does not match ``'[A-Za-z0-9/._-]+'``
 
         """
+        if not self.opened:
+            raise RuntimeError(backend_not_opened_error)
         path = utils.check_path(path)
         return utils.call_function_on_backend(
             self._checksum,
@@ -131,7 +139,9 @@ class Base:
             BackendError: if an error is raised on the backend
 
         """
-        utils.call_function_on_backend(self._close)
+        if self.opened:
+            utils.call_function_on_backend(self._close)
+            self.opened = False
 
     def _copy_file(
         self,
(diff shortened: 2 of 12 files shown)


hagenw:
In this merge request, I just focused to update the `Artifactory` backend to work with the changes proposed here. There are better ways to use `open()` in the `Artifactory` backend, but I will update those in a follow up merge request.

audbackend/core/backend/base.py
```diff
@@ -781,6 +817,7 @@ def open(
 
         """
         utils.call_function_on_backend(self._open)
```
frankenjoe, line 819: Maybe we check if a connection was already established?

audbackend/core/backend/base.py
```diff
@@ -781,6 +817,7 @@ def open(
...
+        self.opened = True
```
frankenjoe, line 820: I wonder if we should rather add a return value to `self._open()` and assign the result here. This avoids that we set the flag when actually no connection could be established.

audbackend/core/backend/base.py
```diff
@@ -131,7 +139,9 @@ def close(
             BackendError: if an error is raised on the backend
 
         """
-        utils.call_function_on_backend(self._close)
+        if self.opened:
+            utils.call_function_on_backend(self._close)
+            self.opened = False
```
frankenjoe, line 144: Here I would also recommend we a return value to `self._close()`, which should be `True` on success and then assign `self.opened = not utils.call_function_on_backend(self._close)`.

audbackend/core/interface/unversioned.py
```diff
@@ -47,6 +48,7 @@ def checksum(
 
         ..
             >>> backend = audbackend.backend.FileSystem("host", "repo")
+            >>> backend.open()
```
frankenjoe, line 51: Maybe use a `with` statement instead? This will safe one line and also makes sure `close()` is called.

audbackend/core/backend/artifactory.py
```diff
@@ -220,7 +208,9 @@ def _delete(
         self,
     ):
         r"""Delete repository and all its content."""
+        self.open()
```
frankenjoe, line 211: I think it's safer to use a `with` statement here to make sure `close()` is still called when the repository cannot be deleted.

docs/developer-guide.rst
```diff
@@ -128,6 +128,7 @@ and upload a file:
 
     audbackend.backend.FileSystem.create("./host", "repo")
     backend = audbackend.backend.FileSystem("./host", "repo")
+    backend.open()
```
frankenjoe, line 131: I think for good practice we should call `close()` before we delete the repository.

audbackend/core/backend/base.py
```diff
@@ -9,6 +9,9 @@
 from audbackend.core.errors import BackendError
 
 
+backend_not_opened_error = "'Backend.open()' needs to be run first."
```
frankenjoe, line 12: Maybe be a little more verbose here, e.g.

```suggestion
backend_not_opened_error = "Call 'Backend.open()' to open a connection to the repository first."
```

audbackend/core/interface/unversioned.py
hagenw: But if I use a `with` statement here, don't I have to indent all the following lines? This I would like to avoid for the examples.

audbackend/core/backend/base.py
hagenw: I see the point, but I'm not sure if it will work out in reality.
Inside `self._open()` it can happen that an error is raised when no connection can be established. If everything went fine, it will return `True` at the end, but when will it ever return `False`?

audbackend/core/backend/base.py
hagenw: Changed this and the corresponding test accordingly.

docs/developer-guide.rst
hagenw: I added:

image

audbackend/core/backend/artifactory.py
hagenw: I changed to
```python
        with self:
            self._repo.delete()
```

audbackend/core/backend/base.py
hagenw: I changed to:
```python
        if not self.opened:
            utils.call_function_on_backend(self._open)
            self.opened = True
```

audbackend/core/interface/unversioned.py
frankenjoe: Why?

audbackend/core/backend/base.py
frankenjoe: Instead of raising an error, the backend could return `False` if no connection could be established.

audbackend/core/interface/unversioned.py
hagenw: Does the `with` statement not always work like this:

```python
with audbackend.backend.FileSystem("host", "repo") as backend:
    backend.put_file(...)
```

How can I use it to achieve:
```python
>>> with audbackend.backend.FileSystem("host", "repo") as backend
>>> backend.put_file(...)
```
?

hagenw: Or are you referring to why I don't want to indent the following lines?
Because the part where we open the backend is at the moment not visible to a reader of the documentation.
We could of course change this, but I know from `auglib` that users are very afraid of using something that contains a `with` statement ;)

audbackend/core/backend/base.py
hagenw: But when not raising an error, the user has no clue what goes wrong.

audbackend/core/interface/unversioned.py
frankenjoe: Ah sorry, I overlooked that these lines are executed **before** the docstring example and since it is anyway not visible to the user we can of course stay with your solution. Only where we show code to the user we should either use a `with` statement or also call `close()`.

hagenw: OK, makes sense, then let's stay with the current solution here.

audbackend/core/backend/base.py
frankenjoe: We could raise an error in the base class if we want to. But I guess you will argument that it's better if the backend raises the error to show a more specific error message. So maybe we should then document somewhere that an error should be raised if the implementation cannot establish a connection or close it.

audbackend/core/interface/unversioned.py
frankenjoe: > But I know from auglib that users are very afraid of using something that contains a with statement

I must admit that I like `with` statements :)

audbackend/core/backend/base.py
hagenw: I adjusted the docstrings to mention that an error should be raised:

```python
    def _close(
        self,
    ):  # pragma: no cover
        r"""Close connection to repository.

        An error should be raised,
        if the connection to the backend
        cannot be closed.

        """
        pass

# ...

    def _open(
        self,
    ):  # pragma: no cover
        r"""Open connection to backend.

        If repository does not exist,
        or the backend cannot be opened,
        an error should be raised.

        """
        pass
```

hagenw:
In the docstring of `audbackend.core.backend.base.Base._close()`, and also when chatting about this topic, we talk about connection to the backend. I was also wondering if I should use `self.connected` instead of `self.opened()`, but I had the impression that connection seems to somehow indicate that the backend is not on the same PC as the user.
Any preference on the naming?

frankenjoe:
> Any preference on the naming?

Yes, `self.connected` sounds better to me. Though we lose a bit the connection to `open()`...

hagenw:
Yes, there is no perfect solution. We could also replace `open()` with `connect()`, but I would not be in favor.
So maybe just stay with `self.opened`.

audbackend/core/backend/base.py
hagenw: Here, I would leave it symmetrical to `self._open()` and also not introduce a return value.

frankenjoe:
Yes, I think `self.opened` is fine.

audbackend/core/backend/base.py
```diff
@@ -25,6 +30,8 @@ def __init__(
         r"""Host path."""
         self.repository = repository
         r"""Repository name."""
+        self.opened = False
+        r"""If backend is opened."""
```
frankenjoe, line 34: Shouldn't it be "is open"?

hagenw: I think I phrased it like this, as I would have also said "is connected".
But "is open" or "was opened" seem to indeed better fit.

I would propose

```suggestion
        r"""If backend is open."""
```

frankenjoe: If we want to be verbose we could also say "If a connection to the repository has been established."

hagenw: Good idea, I changed it accordingly, and updated the screenshot in the description.

frankenjoe: Should we then also change the error description from "RuntimeError: if backend was not opened" to "RuntimeError: if no connection to the repository has been established"?

hagenw: I would stay with the current message, as the connection that the user has to call `backend.open()` is more obvious.

PR #222, opened 2024-04-23
Implement Artifactory._close()

hagenw:
In https://github.com/audeering/audbackend/pull/204 we introduced the backend methods `open()` and `close()` that can be used to establish a connection to a backend. Those are of course also good candidates for handling authentication, if needed by a backend, which we added for `audbackend.backend.Artifactory` in https://github.com/audeering/audbackend/pull/215.

In `audbackend.backend.Artifactory.open()` we store an `artifactory.ArtifactoryPath` object as `audbackend.backend.Artifactory._repo`, which is an opened connection to the repository. But `artifactory.ArtifactoryPath` does not provide a method to close the connection (maybe this is also not needed?).
In the corresponding documentation it is stated that the connection can also be handled by a `requests.Session`, see https://github.com/devopshq/artifactory?tab=readme-ov-file#session, which has the advantage that it provides a `close()` method.

In this pull request, I make use of `requests.Session.close()` and add `audbackend.backend.Artifactory._close()` that calls it to terminate the connection to the Artifactory repository. The session object is stored as `audbackend.backend.Artifactory._session`. I benchmarked the changes when loading emodb as in https://github.com/audeering/audb/issues/387#issuecomment-2071636778, and found no difference in performance.

/cc @frankenjoe

### DIFF
--- a/audbackend/core/backend/artifactory.py
+++ b/audbackend/core/backend/artifactory.py
@@ -178,7 +178,7 @@ class Artifactory(Base):
 
     def _close(
         self,
-    ):  # pragma: no cover
+    ):
         r"""Close connection to repository.
 
         An error should be raised,
--- a/audbackend/core/backend/artifactory.py
+++ b/audbackend/core/backend/artifactory.py
@@ -3,6 +3,7 @@ import typing
 
 import artifactory
 import dohq_artifactory
+import requests
 
 import audeer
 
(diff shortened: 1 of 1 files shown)


ChristianGeng (approved):
Using the session objects from `requests` is imo a good thing to do, as well is implementing a method to close the connection.

I initially had   question regarding the authentication, and asked myself why the authentication data are written into `self.authentication` at all, and whether  could  not simply write `self._session.auth = self.get_authentication(self.host)` or so?  But after quickly looking through the code in the requests library, this does not buy any security advantages and is a non-issue.

I guess there is also a reason for having `self.authentication` and `get_authentication` as public? I guess there is and approve tentatively.

hagenw:
> I guess there is also a reason for having `self.authentication` and `get_authentication` as public? I guess there is and approve tentatively.

Yes, privacy was not one of the design goals so far, as the user that provides the authentication has anyway access to it at the moment. The advantage of having `get_authentication()` is that you can easily see what the current username, password combination is you would get from your config files, environment variables.

PR #231, opened 2024-09-03
Add audbackend.backend.Minio

hagenw:
Implements a backend for [MinIO](https://min.io).

MinIO is compatible with S3 storage, and we tested internally that this backend works with S3 storage on Hetzner and AWS. A test based on AWS is integrated in https://github.com/audeering/audb/pull/450.

I decided to handle the authentication the same way as we do for `audbackend.backend.Artifactory`, by allowing to specify the username / password inside a config file or environment variable. In addition, we have the class method `audbackend.backend.Minio.get_authentication()` to retrieve them, in order to check their values.
For local MinIO servers we also need to set the `secure` argument to `False`. I added this to the config file as well, in oder to make it part of the backend configuration, and added the class method `audbackend.backend.Minio.get_config()`, which returns all entries as dictionary.
In certain edge cases you might want to add other arguments to the underlying `minio.Minio` class. For that reason I decided to add support for `**kwargs`, but do no checks if there is any conflict with authentication as I think this is only for power users anyway.

image

image

image

## Summary by Sourcery

Add support for MinIO as a new backend in the audbackend library, enabling storage operations with MinIO. Refactor tests to accommodate the new backend and ensure robust testing of its functionality.

New Features:
- Introduce a new backend for MinIO, allowing integration with MinIO storage systems.

Enhancements:
- Refactor test parameterization to use a centralized list of backend-interface combinations, improving test maintainability.

Tests:
- Add comprehensive tests for the new MinIO backend, including authentication, file operations, and configuration parsing.

### DIFF
--- a/audbackend/core/backend/minio.py
+++ b/audbackend/core/backend/minio.py
@@ -180,6 +180,9 @@ class Minio(Base):
     ) -> bool:
         r"""Check if file exists on backend."""
         path = self.path(path)
+        if path == "":
+            # Return True for root
+            return True
         try:
             self._client.stat_object(self.repository, path)
         except minio.error.S3Error:
--- a/audbackend/core/backend/minio.py
+++ b/audbackend/core/backend/minio.py
@@ -138,7 +138,7 @@ class Minio(Base):
         src_path = self.path(src_path)
         dst_path = self.path(dst_path)
         # `copy_object()` has a maximum size limit of 5GB.
-        if self._size(src_path) / 1024 / 1024 / 1024 > 5:
+        if self._size(src_path) / 1024 / 1024 / 1024 >= 5:
             with tempfile.TemporaryDirectory() as tmp_dir:
                 tmp_path = audeer.path(tmp_dir, os.path.basename(src_path))
                 self._get_file(src_path, tmp_path, verbose)
--- a/audbackend/core/backend/minio.py
+++ b/audbackend/core/backend/minio.py
@@ -288,8 +288,8 @@ class Minio(Base):
                             dst_fp.write(data)
                             dst_size += n_data
                             pbar.update(n_data)
-            except Exception as e:  # pragma: nocover
-                raise RuntimeError(f"Error downloading file: {e}")  # pragma: nocover
+            except Exception as e:  # pragma: no cover
+                raise RuntimeError(f"Error downloading file: {e}")
             finally:
                 response.close()
                 response.release_conn()
--- a/audbackend/core/backend/minio.py
+++ b/audbackend/core/backend/minio.py
@@ -30,8 +30,10 @@ class Minio(Base):
             when using TLS for the connection,
             and ``False`` otherwise,
             e.g. when using a `local MinIO server`_.
+        **kwargs: keyword arguments passed on to `minio.Minio`_
 
     .. _local MinIO server: https://min.io/docs/minio/container/index.html
+    .. _minio.Minio: https://min.io/docs/minio/linux/developers/python/API.html
 
     Examples:
         >>> host = "play.min.io"  # playground provided by https://min.io
--- a/audbackend/core/backend/minio.py
+++ b/audbackend/core/backend/minio.py
@@ -1,4 +1,5 @@
 import configparser
+import mimetypes
 import os
 import tempfile
 import typing
@@ -369,7 +370,13 @@ class Minio(Base):
             )
             print(desc, end="\r")
 
-        self._client.fput_object(self.repository, dst_path, src_path)
+        content_type = mimetypes.guess_type(src_path)[0] or "application/octet-stream"
+        self._client.fput_object(
+            self.repository,
+            dst_path,
+            src_path,
+            content_type=content_type,
+        )
 
         if verbose:  # pragma: no cover
             # Clear progress line
--- a/audbackend/core/backend/minio.py
+++ b/audbackend/core/backend/minio.py
@@ -1,5 +1,6 @@
 import configparser
 import os
+import tempfile
 import typing
 
 import minio
@@ -133,16 +134,18 @@ class Minio(Base):
         r"""Copy file on backend."""
         src_path = self.path(src_path)
         dst_path = self.path(dst_path)
-        # This has a maximum size limit of 5GB.
-        stats = self._client.stat_object(self.repository, src_path)
-        if stats.size / 1024 / 1024 / 1024 > 5:
-            # TODO: implement copying with temporary file
-            pass
+        # `copy_object()` has a maximum size limit of 5GB.
+        if self._size(src_path) / 1024 / 1024 / 1024 > 5:
+            with tempfile.TemporaryDirectory() as tmp_dir:
+                tmp_path = audeer.path(tmp_dir, os.path.basename(src_path))
+                self._get_file(src_path, tmp_path, verbose)
+                checksum = self._checksum(src_path)
+                self._put_file(tmp_path, dst_path, checksum, verbose)
         else:
             self._client.copy_object(
                 self.repository,
-                src_path,
-                minio.commonconfig.CopySource(self.repository, dst_path),
+                dst_path,
+                minio.commonconfig.CopySource(self.repository, src_path),
             )
 
     def _create(
@@ -159,7 +162,7 @@ class Minio(Base):
     ) -> str:
         r"""Get last modification date of file on backend."""
         path = self.path(path)
-        date = self.client.stat_object(self.repository, path).last_modified
+        date = self._client.stat_object(self.repository, path).last_modified
         date = utils.date_format(date)
         return date
 
@@ -180,9 +183,6 @@ class Minio(Base):
     ) -> bool:
         r"""Check if file exists on backend."""
         path = self.path(path)
-        if path == "":
-            # Return True for root
-            return True
         try:
             self._client.stat_object(self.repository, path)
         except minio.error.S3Error:
@@ -198,7 +198,7 @@ class Minio(Base):
         r"""Get file from backend."""
         src_path = self.path(src_path)
         src_size = self._client.stat_object(self.repository, src_path).size
-        chunk = (4 * 1024,)
+        chunk = 4 * 1024
         with audeer.progress_bar(total=src_size, disable=not verbose) as pbar:
             desc = audeer.format_display_message(
                 "Download {}".format(os.path.basename(str(src_path))),
@@ -209,7 +209,7 @@ class Minio(Base):
 
             dst_size = 0
             try:
-                response = self.client.get_object(self.repository, src_path)
+                response = self._client.get_object(self.repository, src_path)
                 with open(dst_path, "wb") as dst_fp:
                     while src_size > dst_size:
                         data = response.read(chunk)
@@ -244,11 +244,8 @@ class Minio(Base):
         verbose: bool,
     ):
         r"""Move file on backend."""
-        src_path = self.path(src_path)
-        dst_path = self.path(dst_path)
-        if not dst_path.parent.exists():
-            dst_path.parent.mkdir()
-        src_path.move(dst_path)
+        self._copy_file(src_path, dst_path, verbose)
+        self._remove_file(src_path)
 
     def _open(
         self,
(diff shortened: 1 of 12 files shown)


audbackend/core/backend/minio.py
```diff
@@ -0,0 +1,402 @@
...
+
+        Args:
+            path: path on backend
+
+        Returns:
+            path
+
+        """
+        path = path.replace(self.sep, "/")
+        if path.startswith("/"):
```
ChristianGeng, line 354: In Python versions greater than 3.8 it would be safe to remove the conditional and replace it with. `path = path.removeprefix("/")`.  I believe that it is getting close to deprecating 3.8?

audbackend/core/backend/minio.py
```diff
@@ -0,0 +1,402 @@
...
+    def _get_file(
+        self,
+        src_path: str,
+        dst_path: str,
+        verbose: bool,
+    ):
+        r"""Get file from backend."""
+        src_path = self.path(src_path)
+        src_size = self._client.stat_object(self.repository, src_path).size
+        chunk = 4 * 1024
```
ChristianGeng, line 270: These are  4 KB. Is this chunk size motivated / tested?  Or ad hoc?

audbackend/core/backend/minio.py
```diff
@@ -0,0 +1,402 @@
...
+        self,
+    ):
+        r"""Close connection to repository.
+
+        An error should be raised,
+        if the connection to the backend
+        cannot be closed.
+
+        """
+        # At the moment, this is automatically handled.
```
ChristianGeng, line 180: Automatically handled means by the minio client lib. Would it make sense to make this explicit?

audbackend/core/backend/minio.py
```diff
@@ -0,0 +1,402 @@
...
+    def _copy_file(
+        self,
+        src_path: str,
+        dst_path: str,
+        verbose: bool,
+    ):
+        r"""Copy file on backend."""
+        src_path = self.path(src_path)
+        dst_path = self.path(dst_path)
+        # `copy_object()` has a maximum size limit of 5GB.
```
ChristianGeng, line 206: The 5GB limit is a limit of S3 native? [This post](https://stackoverflow.com/questions/33356041/technically-what-is-the-difference-between-s3n-s3a-and-s3) suggests this at least.  So the minio client uses s3n?

audbackend/core/backend/minio.py
hagenw: When removing Python 3.8, there are other things we should update, e.g. if I remember correctly the `typing` module might no longer be needed. I would prefer to first create a general issue to gather all the things we would like to update when removing Python 3.8, and then do this in dedicated pull requests.

Which means for now in this pull request we need to still support Python 3.8.

audbackend/core/backend/minio.py
hagenw: I just copied it from the Artifactory backend.

@frankenjoe did you ever tested the influence if the chunk size for downloading data, or did you (or I) copied this from another package?

audbackend/core/backend/minio.py
hagenw: Don't know, I just found the information in the MinIO documentation. As we don't have a test for this, maybe we should also lower the value slightly to be on the safe side?

audbackend/core/backend/minio.py
```diff
@@ -0,0 +1,402 @@
...
+        path: str,
+    ):
+        r"""Remove file from backend."""
+        path = self.path(path)
+        self._client.remove_object(self.repository, path)
+
+    def _size(
+        self,
+        path: str,
+    ) -> str:
```
ChristianGeng, line 405: This is used by `_copy_file` where it is used to determine whether file size exceeds 5GB.  However I can see no cast in `_copy_file`. Should the return type hint be integer?

tests/test_backend_minio.py
```diff
@@ -0,0 +1,319 @@
...
+@pytest.mark.parametrize("host", [pytest.HOSTS["minio"]])
+@pytest.mark.parametrize("repository", [f"unittest-{pytest.UID}-{audeer.uid()[:8]}"])
+@pytest.mark.parametrize("authentication", [("bad-access", "bad-secret")])
+def test_errors(host, repository, authentication):
+    backend = audbackend.backend.Minio(host, repository, authentication=authentication)
+    with pytest.raises(audbackend.BackendError):
+        backend.open()
+
+
+def test_get_config(tmpdir, hosts, hide_credentials):
```
ChristianGeng, line 154: This tests parsing of the configuration, but does not have a one-line docstring.

tests/test_backend_minio.py
```diff
@@ -0,0 +1,319 @@
...
+            "1.0.0",
+            [r"(\d+.)?tar.gz"],
+            True,
+            "/file/1.0.0/file-1.0.0.tar.gz",
+        ),
+    ],
+)
+def test_maven_file_structure(
+    tmpdir, interface, file, version, extensions, regex, expected
+):
```
ChristianGeng, line 306: Although the test name is already telling, a single line docstring would be nice.

audbackend/core/backend/minio.py
frankenjoe: No, I don't think we ever tested a different chunk size.

hagenw: Thanks for the feedback.

I would propose to investigate this at a later stage and independent from this pull request. I created https://github.com/audeering/audbackend/issues/235 to track it.

tests/test_interface_unversioned.py
```diff
@@ -91,6 +91,7 @@ def tree(tmpdir, request):
     [
         (audbackend.backend.Artifactory, audbackend.interface.Unversioned),
         (audbackend.backend.FileSystem, audbackend.interface.Unversioned),
+        (audbackend.backend.Minio, audbackend.interface.Unversioned),
```
ChristianGeng, line 94: The interface parametrization seems to vary little between tests. 

Would it make sense to recast this into a fixture?  Possibly like this:


```python
@pytest.fixture(params=[
    (audbackend.backend.Artifactory, audbackend.interface.Unversioned),
    (audbackend.backend.FileSystem, audbackend.interface.Unversioned),
    (audbackend.backend.Minio, audbackend.interface.Unversioned),
    (SingleFolder, audbackend.interface.Unversioned),
])
def interface(request):
    return request.param
```

audbackend/core/backend/minio.py
ChristianGeng: Probably lowering would be ok to get it working right now. 
The background is a different  one: if MinIO has the limit of 5GB this suggests that the minio package  is using the old s3n interface.  Afaiu s3a succeeds s3n. The difference between is that s3n supports objects up to 5GB, but s3a should support 5TB and has improved performance. I cannof find the link right now, but I think I had seen that `fsspec` has an implementation of s3a.

tests/test_backend_minio.py
hagenw: I added a docstring.

tests/test_backend_minio.py
hagenw: I added a docstring.

tests/test_interface_unversioned.py
hagenw: `interface` was already a fixture, but I could solve the problem by a variable:

```python
# Backend-interface combinations to use in all tests
backend_interface_combinations = [ 
    (audbackend.backend.Artifactory, audbackend.interface.Unversioned),
    (audbackend.backend.FileSystem, audbackend.interface.Unversioned),
    (audbackend.backend.Minio, audbackend.interface.Unversioned),
    (SingleFolder, audbackend.interface.Unversioned),
]

# ...

@pytest.mark.parametrize(
    "interface",
    backend_interface_combinations,
    indirect=True,
)
def test_archive(tmpdir, tree, archive, files, tmp_root, interface, expected):
```

audbackend/core/backend/minio.py
hagenw: I removed `_close()` and added `close()` with an updated docstring instead:

image

hagenw: In principle, we can also take care about opening and closing by providing an `http_client` object to MinIO, but as it will be very complicated to find out the best settings in that case, I would stay for now with the default and just state that closing is automatically handled by MinIO.

audbackend/core/backend/minio.py
hagenw: Good catch, it returns indeed an integer. I corrected the typing information.

audbackend/core/backend/minio.py
hagenw: This is now tracked in [tools/meta#12] internally.

audbackend/core/backend/minio.py
```diff
@@ -0,0 +1,423 @@
...
+
+        Returns:
+            path
+
+        """
+        path = path.replace(self.sep, "/")
+        if path.startswith("/"):
+            # /path -> path
+            path = path[1:]
+        return path
```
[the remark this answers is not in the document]
hagenw: As we still support Python 3.8 here, we will stay with the current solution.

audbackend/core/backend/minio.py
```diff
@@ -0,0 +1,423 @@
...
+        src_path: str,
+        dst_path: str,
+        verbose: bool,
+    ):
+        r"""Copy file on backend."""
+        src_path = self.path(src_path)
+        dst_path = self.path(dst_path)
+        # `copy_object()` has a maximum size limit of 5GB.
+        # We use 4.9GB to have some headroom
+        if self._size(src_path) / 1024 / 1024 / 1024 >= 4.9:
```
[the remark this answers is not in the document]
hagenw: This does not improve the code.

tests/test_backend_minio.py
```diff
@@ -0,0 +1,345 @@
...
+    with open(config_path, "w") as fp:
+        fp.write(f"[{host}.abc]\n")
+    config = audbackend.backend.Minio.get_config(host)
+    assert config == {}
+
+    # config file entry without variables
+    with open(config_path, "w") as fp:
+        fp.write(f"[{host}]\n")
+    config = audbackend.backend.Minio.get_config(host)
+    assert config == {}
```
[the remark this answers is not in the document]
hagenw: To make the tests more comprehensible at first glance, I will not update here.

audbackend/core/backend/minio.py
```diff
@@ -0,0 +1,423 @@
...
+                    while src_size > dst_size:
+                        data = response.read(chunk)
+                        n_data = len(data)
+                        if n_data > 0:
+                            dst_fp.write(data)
+                            dst_size += n_data
+                            pbar.update(n_data)
+            finally:
+                response.close()
+                response.release_conn()
```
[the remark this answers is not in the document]
hagenw: I added the suggested exception.

tests/conftest.py
```diff
@@ -15,6 +15,35 @@
...
+    """Provide authentication tokens for supported backends."""
+    if pytest.HOSTS["minio"] == "play.min.io":
+        defaults = {}
+        for key in [
+            "MINIO_ACCESS_KEY",
+            "MINIO_SECRET_KEY",
+        ]:
+            defaults[key] = os.environ.get(key, None)
+
+        os.environ["MINIO_ACCESS_KEY"] = "Q3AM3UQ867SPQQA43P2F"
```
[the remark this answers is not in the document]
hagenw: The access keys are public in the MinIO docs and desired to test code. I added a comment to make it more imlicit, that this is not an issue here.

audbackend/core/backend/minio.py
```diff
@@ -0,0 +1,423 @@
...
+    ):
+        r"""Open connection to backend."""
+        # At the moment, session management is handled automatically.
+        # If we want to manage this ourselves,
+        # we need to use the `http_client` argument of `minio.Minio`
+        if not self._client.bucket_exists(self.repository):
+            utils.raise_file_not_found_error(self.repository)
+
+    def _owner(
+        self,
```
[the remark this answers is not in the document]
hagenw: I updated the code accordingly.

tests/conftest.py
```diff
@@ -15,6 +15,35 @@
...
 
+# Define static hosts
+pytest.HOSTS = {
+    "artifactory": "https://audeering.jfrog.io/artifactory",
+    "minio": "play.min.io",
+}
+
+
+@pytest.fixture(scope="package", autouse=True)
+def authentication():
```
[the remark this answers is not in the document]
hagenw: I updated the code accordingly.

hagenw:
I addressed now all comments and updated the description by mentioning our Hetzner S3 test.

hagenw:
As it works with `audb` (see https://github.com/audeering/audb/pull/450), I will merge this in order to continue with related updates.

PR #234, opened 2024-09-30
Add support for owner()

hagenw:
This is not based on `main`, but expands the `minio` branch.

MinIO seems not directly to support the concept of a file owner. At least if you call their `stats_object()` method and look at its attribute `owner_name` it is always empty for me.

This pull request adds a workaround, by setting the owner explicitly as a metadata entry, every time a file is uploaded or copied to the server. The owner can then be extracted with `stats_object().metadata`. I benchmarked, and the proposed changes have no affect on the execution time of `put_file()`.

---

We should ask ourselves, if we really want to have this feature or not. If we later switch to `fsspec` for backends (https://github.com/audeering/audbackend/pull/233), we will again have the problem, that `owner` is not supported, and we will most likely not be able to add a workaround that is independent from the backend. So maybe we should just drop it?

The challenge is that it provides very meaningful information, e.g. who published a given model or data, which means if we drop it from `audbackend`, we will most likely need to store those information inside other metadata, e.g. in the header of a dataset.

### DIFF
--- a/audbackend/core/backend/minio.py
+++ b/audbackend/core/backend/minio.py
@@ -1,4 +1,5 @@
 import configparser
+import getpass
 import mimetypes
 import os
 import tempfile
@@ -213,6 +214,7 @@ class Minio(Base):
                 self.repository,
                 dst_path,
                 minio.commonconfig.CopySource(self.repository, src_path),
+                metadata=_metadata(),
             )
 
     def _create(
@@ -330,9 +332,11 @@ class Minio(Base):
     ) -> str:
         r"""Get owner of file on backend."""
         path = self.path(path)
-        # TODO: owner seems to be empty,
-        # need to check if we have to manage this ourselves?
-        owner = self._client.stat_object(self.repository, path).owner_name
+        # NOTE:
+        # we use a custom metadata entry to track the owner
+        # as stats.owner_name is always empty.
+        stats = self._client.stat_object(self.repository, path)
+        owner = stats.metadata["x-amz-meta-owner"]
         return owner
 
     def path(
@@ -376,6 +380,7 @@ class Minio(Base):
             dst_path,
             src_path,
             content_type=content_type,
+            metadata=_metadata(),
         )
 
         if verbose:  # pragma: no cover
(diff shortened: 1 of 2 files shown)


hagenw:
I would vote to support `owner()` in the `MinIO` backend (https://github.com/audeering/audbackend/pull/231) we introduce with the next release of `audbackend`.

We can then start in parallel to update our packages like `audb` and `audmodel`, to no longer rely on `audbackend.Interface.Base.owner()`, but store the information in their own metadata. When we then target `fsspec` support (https://github.com/audeering/audbackend/pull/233) in the next mayor release of `audbackend`, we can then drop `owner()` in `audbackend`.

audbackend/core/backend/minio.py
```diff
@@ -398,3 +405,13 @@ def _size(
         path = self.path(path)
         size = self._client.stat_object(self.repository, path).size
         return size
+
+
+def _metadata():
```
ChristianGeng, line 410: Is there a reason to maintain `_metadata` completely separate from the `Minio` class? 
Intuitively one would think of `_metadata` as a (private) property. On the other hand, the class needs to implement the interface methods from the parent class, and a property might become confusing.

ChristianGeng (approved):
# Problem solved in this MR

This solves the problem that in the current state of our minio deployments the `owner_name` is always empty. The particular workaround applied fetches the owner attribute from the from the minio client itself.

This is a good working option in the present scenario. The owner field in `self._client.stat_object(self.repository, path).owner_name` is empty when using minio's internal access management. In case that that a switch to an oidc external identity provider the owner field might contain values that then would become the preferrable option, but for now the implementation allows to track the user in a sufficient fashion.

So in sum I think that conceptually this is ready for approval.

audbackend/core/backend/minio.py
hagenw: I tend to implement such helper functions independent of the class. My main reason is that in the class we only have the methods that needed to be added, and it does not get cluttered. But maybe it has also some drawbacks?

PR #244, opened 2024-11-13
Store checksum in metadata on S3/MinIO

hagenw:
Tackles part of https://github.com/audeering/audb/issues/459

So far we assumed that the `etag` of an object stored in MinIO or S3 is the MD5 sum of the uploaded file. But this is not the case for larger files, as mentioned at https://docs.aws.amazon.com/AmazonS3/latest/API/API_Object.html.

To solve the above issue, this pull request stores a given checksum as a metadata entry on S3 and MinIO backends.

This is also in line with the Artifactory backend, where we also provide the checksum to be stored on the server (directly in the put method as Artifactory supports that):

https://github.com/audeering/audbackend/blob/e75620a975006ef2a57c031a6f87a84cc91c6049/audbackend/core/backend/artifactory.py#L32
In addition, in order to solve https://github.com/audeering/audb/issues/459 for parquet files, we will need an option to store a manual provided value on the server, which is also enabled by the changes presented here.

## Summary by Sourcery

Enhancements:
- Store checksum in metadata for files on S3/MinIO to ensure accurate file verification.

### DIFF
--- a/audbackend/core/backend/minio.py
+++ b/audbackend/core/backend/minio.py
@@ -177,7 +177,8 @@ class Minio(Base):
     ) -> str:
         r"""MD5 checksum of file on backend."""
         path = self.path(path)
-        return self._client.stat_object(self.repository, path).etag
+        meta = self._client.stat_object(self.repository, path).metadata
+        return meta["x-amz-meta-checksum"] if "x-amz-meta-checksum" in meta else ""
 
     def _collapse(
         self,
@@ -202,20 +203,20 @@ class Minio(Base):
         r"""Copy file on backend."""
         src_path = self.path(src_path)
         dst_path = self.path(dst_path)
+        checksum = self._checksum(src_path)
         # `copy_object()` has a maximum size limit of 5GB.
         # We use 4.9GB to have some headroom
         if self._size(src_path) / 1024 / 1024 / 1024 >= 4.9:
             with tempfile.TemporaryDirectory() as tmp_dir:
                 tmp_path = audeer.path(tmp_dir, os.path.basename(src_path))
                 self._get_file(src_path, tmp_path, verbose)
-                checksum = self._checksum(src_path)
                 self._put_file(tmp_path, dst_path, checksum, verbose)
         else:
             self._client.copy_object(
                 self.repository,
                 dst_path,
                 minio.commonconfig.CopySource(self.repository, src_path),
-                metadata=_metadata(),
+                metadata=_metadata(checksum),
             )
 
     def _create(
@@ -375,7 +376,7 @@ class Minio(Base):
             dst_path,
             src_path,
             content_type=content_type,
-            metadata=_metadata(),
+            metadata=_metadata(checksum),
         )
 
         if verbose:  # pragma: no cover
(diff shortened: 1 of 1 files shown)


audbackend/core/backend/minio.py
```diff
@@ -202,20 +203,20 @@ def _copy_file(
         r"""Copy file on backend."""
         src_path = self.path(src_path)
         dst_path = self.path(dst_path)
+        checksum = self._checksum(src_path)
         # `copy_object()` has a maximum size limit of 5GB.
         # We use 4.9GB to have some headroom
         if self._size(src_path) / 1024 / 1024 / 1024 >= 4.9:
```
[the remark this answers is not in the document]
hagenw: This is unrelated to this pull request and should stay as it is for easier readability.

audbackend/core/backend/minio.py
```diff
@@ -177,7 +177,8 @@ def _checksum(
     ) -> str:
         r"""MD5 checksum of file on backend."""
         path = self.path(path)
-        return self._client.stat_object(self.repository, path).etag
+        meta = self._client.stat_object(self.repository, path).metadata
+        return meta["x-amz-meta-checksum"] if "x-amz-meta-checksum" in meta else ""
```
ChristianGeng, line 181: I am  thinking about the meaning of `if "x-amz-meta-checksum" in meta else` . 
What are the conditions in which `metadata` does not contain the `x-amz-meta-checksum` field? 

I see that it is retrieved by `self._client.stat_object`, and this can be calculated on local files (as in ` self._checksum(src_path)` in the `_copy_file` method. 

So my assumption would have been that it is always there, right?

hagenw: Yes, `x-amz-meta-` is always there, see https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html#UserMetadata

ChristianGeng: But why then use `if "x-amz-meta-checksum" in meta else ""`?

ChristianGeng: I see.  I have a problem overseeing what could be the worst case that happens by a missing (empty) checksum.

ChristianGeng: This is what the AI says when running it locally: 

```
suggestion(edge_case_not_handled): Consider handling missing checksum more explicitly.
Returning an empty string when the checksum is missing might lead to silent failures or unexpected behavior. Consider raising an exception or logging a warning to make it clear that the checksum is not available.

```

hagenw: Thanks for raising this concern, this is indeed worth some additonal thought.

The worst case that can happen is that files will be transferred again (which also happens in the current `main` branch as the wrong MD5 sum is returned for larger files). E.g. if `put_file()` finds the same filename on the server it checks if the checksum match, and if not upload the file again. The checksum match is calculated as a string comparison, so it will not result in an error (which would be the case if we return `None` instead of an empty string).

I would not raise an exception or a warning, as `audbackend` should also work for servers, that contain files, that were uploaded with different software. In that case, they will most likely not have a `x-amz-meta-checksum` entry.
But maybe we should change the names to `x-amz-meta-audbackend-checksum` and `x-amz-meta-audbackend-owner` to avoid colliding with metadata created by other software?

One alternative I see for returning `""` if the metadata entry is missing, is returning the `etag`, which we do currently in `main`. Any preference what to return?

ChristianGeng: >> One alternative I see for returning "" if the metadata entry is missing, is returning the etag, which we do currently in main. Any preference what to return?

This merge request deliberately departs from `etag` as it can become wrong for large files. I would think that the current checksum is an improvement in that sense.

ChristianGeng: Yes I would say so too. The values are there to validate data artifact identity, and the worst case again is copying back and fro data. Therefore resolving.

PR #246, opened 2024-11-14
Remove deprecated functions

hagenw:
In the release 2.0.0 (2024-05-10) of `audbackend` we deprecated the functions `audbackend.access()`, `audbackend.create()`, `audbackend.delete()`, and `audbackend.register()`, and marked them for removal in version 2.2.0. At the moment, we are at version 2.1.0, and the next release will be 2.2.0 as we added a new function in https://github.com/audeering/audbackend/pull/245.

To address it, this pull requests removes:

* `audbackend.access()`
* `audbackend.create()`
* `audbackend.delete()`
* `audbackend.register()`

It adds a few new tests in order to preserve code coverage, as before the tests for the deprecated functions were responsible to cover some parts.

## Summary by Sourcery

Tests:
- Add new tests to ensure code coverage after removing deprecated functions.

## Summary by Sourcery

Tests:
- Add new tests to ensure code coverage after removing deprecated functions.


ChristianGeng (approved):
## Review of #246 

Afaics this completes a process that has been broader in scope than simplz removing  deprecated functions:
together with previous pull requests, the design architecture has changed in that
there now is a division between interface and backend.

The now deleted `audbackend/core/api.py` had been implementing a kind of delegation design pattern.  with public methods being mapped to subclasses' private method by means of a delegation design pattern.

This is now not needed any longer, simplifying the development process.
I sum, I have no low level technical comment beyond sourcery's, but
the change in design has left a few scars in the documentation:

I find that the depiction of the two components:




(i) a backend that implements file operations on a specific storing device (`audbackend.backend`) and (ii) an interface that passes user requests to a backend (`audbackend.interface`) is not optimally located at the beginning of the API doc.

I think it would make more sense to have such a high-level description more upfront, e.g. at the top of developer guide or the usage guide in order to get the documentation reader oriented as fast as possible.

The points that I have raised now do not really target the current Merge Request, rather might stimulate a follow-up issue if at all.  As said, apart from that I do not want to get into low level details that looking are good, at least afaics.

Therefore I am approving right now.

hagenw:
Thanks for your points regarding how to improve the documentation. Would be cool, if you could create an issue for that, so we don't forget it.

ChristianGeng:
> Thanks for your points regarding how to improve the documentation. Would be cool, if you could create an issue for that, so we don't forget it.

see #251

PR #255, opened 2024-11-22
Remove audbackend.checksum() and use MD5 sum

### LINKED ISSUE #254: Artifactory backend refuses to upload parquet files
In https://github.com/audeering/audbackend/pull/245 we introduced `audbackend.checksum()` that calculates the checksum for a parquet file only as MD5 sum, if it is not given in its metadata. In the Artifactory backend, the checksum is then handed to the server inside its deploy function:

https://github.com/audeering/audbackend/blob/1b1177de47dbe4e8eb58f19153ddc6afbd86ccae/audbackend/core/backend/artifactory.py#L31-L32

But the problem is it uses this MD5 sum to check if the data is correct after the upload, which means for parquet files with use checksums that deviate from their MD5 sum, the upload will fail.

Minimal failing example:

```python
import audb
import audeer
import audformat.testing


build_dir = audeer.path("./build")
audeer.rmdir(build_dir)
audeer.mkdir(build_dir)

db = audformat.testing.create_db(minimal=True)
db.name = "age-test"
db.license = "CC0-1.0"
db.schemes["age"] = audformat.Scheme("int", minimum=20, maximum=90)
audformat.testing.add_table(
    db, 
    table_id="age",
    index_type="filewise",
    columns="age",
    num_files=3,
)
db.save(build_dir)
audformat.testing.create_audio_files(db)

repository = audb.Repository(
    "data-public-local",
    "https://artifactory.audeering.com/artifactory",
    "artifactory",
)
audb.publish("./build", "1.0.0", repository, previous_version=None)
```

This fails with:
```
...
dohq_artifactory.exception.ArtifactoryException: Checksum policy 'LocalRepoChecksumPolicy: CLIENT' rejected the artifact 'data-public-local:age-test/meta/age/1.0.0/age-1.0.0.parquet'. Checksums info: ChecksumsInfo{checksums={SHA-1=ChecksumInfo{type=SHA-1, original='null', actual='022a01fc6ec734787d039f3b573929feb80ce603'}, MD5=ChecksumInfo{type=MD5, original='af1eca9bf34749437c01bb0ef1a5b253', actual='e7d53a654dda4aeac899a43e60321617'}, SHA-256=ChecksumInfo{type=SHA-256, original='null', actual='81fee05ddf2bd1b5e571d43cfb137df770c6ae50904252f0b43d486e6f8350d9'}}}
...
audbackend.core.errors.BackendError: An exception was raised by the backend, please see stack trace for further information.
```

hagenw:
Closes #254 

This adds a test for #254 and fixes it by reversing the introduction of `audbackend.checksum()` from https://github.com/audeering/audbackend/pull/245 as we shouldn't upload the checksum stored in the metadata of the parquet file, but as we did before always use the MD5 sum.

This will have no negative effect on version tracking with `audb` (i was wrong when stating this in https://github.com/audeering/audb/issues/459). The only thing affected by uploading the MD5 sum instead of the parquet file hash is: if a user cancels an `audb.publish()` job, deletes the build folder, creates the table files again and starts the upload again, it will overwrite the parquet table files instead of skipping them, as their MD5 sum has changed locally. But this should be fine. If we stay with the current implementation instead, we have no way to verify if a file was uploaded/downloaded correctly as we don't store information on the actual MD5 sum.

## Summary by Sourcery

Add a test for parquet file upload to Artifactory and fix the issue by reverting to using the MD5 checksum instead of the metadata checksum.

Bug Fixes:
- Fix the issue with uploading parquet files by reverting to using the MD5 checksum instead of the checksum stored in the metadata.

Tests:
- Add a test to ensure parquet files are uploaded with the correct MD5 checksum instead of the metadata checksum.

## Summary by Sourcery

Revert the checksum calculation for parquet files to use the MD5 checksum instead of the metadata checksum to fix upload issues. Add a test to verify the correct checksum is used during the upload process.

Bug Fixes:
- Fix the issue with uploading parquet files by reverting to using the MD5 checksum instead of the checksum stored in the metadata.

Tests:
- Add a test to ensure parquet files are uploaded with the correct MD5 checksum instead of the metadata checksum.


audbackend/core/backend/base.py
```diff
@@ -69,12 +69,12 @@ def _assert_equal_checksum(
 
         """
         if path_is_local:
-            checksum = utils.checksum(path)
+            checksum = audeer.md5(path)
```
ChristianGeng, line 72: At first, it felt a little weird seing the identifiers `md5sum` and `checksum` so close together feels.  MD5  is already a type of checksum.  

 `audbackend.checksum` calls the private method `_checkum` of the concrete backend implementation . For example in minio `stat_object.metadata` of the client is fetched, and the checksum is extracted from the "x-amz-meta-checksum". For Artifactory the method is called md5 from the `.stat` field of a path. So checksum can be an md5 but does not have to be because it uses a backend implementation. 

Would it make sense to factor the checksum call into a  private method for pure readability that contains a docstring that contains this information (cross-referencing the backend implementations)?

ChristianGeng: I also find that `_assert_equal_checksum`:
path represents the current file being processed or checked, and 
path_ref represents a reference file against which the first file's checksum is being compared.

I find the information that the function makes no assumption about the `path` and ` path_ref` valuable. In other words both could be local or both could be remote files.  Despite being a private method, would a docstring be a possibility to clarify this?

ChristianGeng: Is the diagram in https://github.com/audeering/audbackend/pull/255#issuecomment-2493135995 correct?
My understanding is that `backend.put_file` triggers the file upload. There is a class `Interface` but it is not relevant for the backend interaction.

hagenw: The former introduced function `audbackend.checksum()` was implemented in `audbackend.core.utils.checksum()`. It was used on files stored at the current local disc of a user (basically, files outside of the backend the files should be uploaded to). This is given by the `if path_is_local` command above. Those cases are now replaced by calling `audeer.md5()` as we did before https://github.com/audeering/audbackend/pull/245.
This also means `audbackend.checksum()` does not call the private `_checkum()` method. This method is instead called by all backend methods deriving from `audbackend.backend.Base.checksum()`, and are meant to get the checksum from a backend (Artifactory, S3, filesystem). For the filesystem backend, it also calls `audeer.md5()`.

> Would it make sense to factor the checksum call into a private method for pure readability that contains a docstring that contains this information (cross-referencing the backend implementations)?

You would vote for still having `audbackend.core.utils.checksum()` (that has a proper docstring and just calls `audeer.md5()`), but not include it in the public API, correct?

> My understanding is that backend.put_file triggers the file upload. There is a class Interface but it is not relevant for the backend interaction.

In the interface classes, the underlying backend methods are called, e.g. `audbackend.interface.Unversioned.checksum()`:

https://github.com/audeering/audbackend/blob/1b1177de47dbe4e8eb58f19153ddc6afbd86ccae/audbackend/core/interface/unversioned.py#L38-L71

ChristianGeng: >> You would vote for still having audbackend.core.utils.checksum() (that has a proper docstring and just calls audeer.md5()), but not include it in the public API, correct?

I am not sure - or not really. My main point is to make using the class and understanding it easy for the developer. MD5s are a kind of checksum.  I find that the conditional


```python
        if path_is_local:
            checksum = audeer.md5(path)
        else:
            checksum = self.checksum(path)
```

might trigger misunderstanding.  And I also find that putting in an intermediate layer that polishes the calling convention is a remedy but not a cure - and I don't have one either. Maybe a docstring would do?

>> In the interface classes, the underlying backend methods are called, e.g. audbackend.interface.Unversioned.checksum():

Ok yes. But what is being calculated comes from the backend, i.e. the backend devices on the  implementation - the interface is transparent.

hagenw: OK, I updated the code to use the term `md5`:

```python
        md5 = audeer.md5(path) if path_is_local else self.checksum(path)
        md5_ref = audeer.md5(path_ref) if path_ref_is_local else self.checksum(path_ref) 
```

and I extended the docstring to:

```python
        r"""Assert checksums are equal.

        Compare the MD5 sum of a file
        (``path``)
        to the MD5 sum of a reference file
        (``path_ref``).
        If check fails,
        ``path`` is removed
        and an error is raised.

        Both ``path`` and ``path_ref``
        can be local files,
        or stored on any backend.

        Args:
            path: path to a file.
                Its MD5 sum is compared
                to a reference one,
                calculated from ``path_ref``
            path_is_local: if ``True``,
                assumes ``path`` is stored on local machine
            path_ref: path to a file.
                Its MD5 sum is used as reference
            path_ref_is_local: if ``True``,
                assumes ``path_ref`` is stored on local machine

        Raises:
            InterruptedError: if the MD5 sums do not match

        """
```

ChristianGeng: Nice.

ChristianGeng (approved):
I only had a minor issue concerning possible developer confusion 
relating the terminology of  `md5sum`  and `checksum`. 

This is resolved as good as it can get. So, approval is being given.

audbackend/core/backend/base.py
hagenw: Thanks, I will merge then.

PR #261, opened 2025-01-23
Release 2.2.2

hagenw:
The fix for Python 3.12 they introduced in `dohq-artifactory` 1.0.0 had some breaking changes how path names are returned (e.g. `/` at the end of an URL was removed), so I had to adjust the code here as well (done in https://github.com/audeering/audbackend/pull/260). I then enabled support for Artifactory Python 3.12 here as well (done in https://github.com/audeering/audbackend/pull/259).

image

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,18 @@ The format is based on `Keep a Changelog`_,
 and this project adheres to `Semantic Versioning`_.
 
 
+Version 2.2.2 (2025-01-23)
+--------------------------
+
+* Added: support for Python 3.13
+  (without Artifactory backend)
+* Added: support for Artifactory backend
+  in Python 3.12
+* Changed: depend on ``dohq-artifactory>=1.0.0``
+* Fixed: wrong path names
+  in Artifactory backend with ``dohq-artifactory==1.0.0``
+
+
 Version 2.2.1 (2024-11-27)
 --------------------------
 


CHANGELOG.rst
```diff
@@ -7,6 +7,18 @@ The format is based on `Keep a Changelog`_,
 and this project adheres to `Semantic Versioning`_.
 
 
+Version 2.2.2 (2025-01-23)
+--------------------------
+
+* Added: support for Python 3.13
```
ChristianGeng, line 13: I wonder whether the information that the Artifactory backend is incompatible with 3.13 should also go to other visible locations -  One could think of `install.rst` or `pyproject.toml`.   Or is that handled in the installation process?

hagenw: In `pyproject.toml` it is also handled:

```toml
[project.optional-dependencies]
artifactory = [ 
    'dohq-artifactory >=1.0.0; python_version < "3.13"',
]
minio = [ 
    'minio',
]
all = [ 
    'dohq-artifactory >=1.0.0; python_version < "3.13"',
    'minio',
]
```

ChristianGeng: Yes it's there - sorry. I overread it.  
Then  I will go ahead with approval quickly.

ChristianGeng (approved):
It can be approved. The issue that I had turned out to be a non-issue.

PR #287, opened 2026-02-23
Use connection pool for faster download

hagenw:
Addresses https://github.com/audeering/audb/issues/546

This uses the proposal from https://github.com/audeering/audb/issues/546 3b and uses pooled connections to avoid having to establish a HTTPS connection for every single file that it will download. Currently, max connection is set to 10 and can be changed in the config file (maybe we should increase default to 100?).

I did a benchmark on a dataset with ~1,500,000 files and the remaining time after 5 minutes into the download with 6 workers is reduced from 175:40:24 hours to 167:05:28 hours. ~~I will provide another benchmark with 48 workers at a later stage.~~ A first test with 48 workers and max connection of 100 did not show any improvement. In both cases the time was ~155 hours.

It also adds the `get_config()` method to `audbackend.backend.Artifactory` to be in line with `audbackend.backend.Minio`. The config file is still stored at a different place, but this will be handled in a different pull request (https://github.com/audeering/audbackend/issues/288).

image

image

image

### DIFF
--- a/audbackend/core/backend/artifactory.py
+++ b/audbackend/core/backend/artifactory.py
@@ -199,7 +199,7 @@ class Artifactory(Base):
 
         .. code-block:: ini
 
-            [https://artifactory.example.com/artifactory]
+            [artifactory.example.com/artifactory]
             pool_connections = 10
             pool_maxsize = 100
             max_retries = 3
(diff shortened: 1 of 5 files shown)


hagenw:
This has not shown a big boost in performance, so I will set it to draft for now.

PR #309, opened 2026-08-05
CI: run minio tests on self-hosted MinIO

ChristianGeng:
Closes #308

## Summary

The minio backend tests depended on `play.min.io`, MinIO's public demo
server, which has become unreachable for a sustained period. This hung
CI for an hour before failing, on every OS/Python combination, unrelated
to whatever else is being changed in a given PR (first surfaced on #307).

Confirmed via curl timeouts from multiple networks, and a traceroute
that shows packets reaching a boundary router near where play.min.io's
IP appears to be hosted before getting no response at all -- consistent
with an outage at or near the destination, not a local network issue.
Regardless of the exact cause, the takeaway is that play.min.io cannot
be relied on for CI.

## Fix

Run a self-hosted, pinned MinIO (`RELEASE.2025-09-07T16-13-09Z`, the
same version on every platform) instead of the public playground:

- **Linux**: a `minio/minio` Docker container, started via a plain
  `docker run` step rather than a GitHub Actions `services:` container
  (`services:` has no way to pass a command, and this image has no
  default CMD).
- **Windows/macOS**: Docker is not pre-installed on those runners, so
  these jobs download the native MinIO server binary from dl.min.io
  (pinned archive URLs, verified to exist) and start it as a background
  process -- `nohup` on macOS, `Start-Process` on Windows. All minio
  tests run on all three platforms.

An `AUDBACKEND_SKIP_MINIO_TESTS` escape hatch remains for local
development on machines without a MinIO server, but CI does not use it.

## Also fixed along the way

- A latent bug in `Minio.__init__`: `secure` read from the config file
  was never actually parsed as a boolean. `configparser` returns string
  values, so `secure = False` in a config file produced the Python
  string `"False"`, which is truthy, silently enabling HTTPS instead of
  plain HTTP. Needed this fixed to make a local, non-TLS MinIO server
  work at all. Added a parametrized regression test covering the
  false-like and true-like string values.
- The class docstring's `Examples` still hit the real `play.min.io`;
  marked the two network-touching lines `# doctest: +SKIP`.

## Test plan

- [x] Verified locally end-to-end against a real `minio/minio` container
      and against the native linux-amd64 binary: full suite passes with
      100% coverage (645 passed), down from the hour-long hang
- [x] CI passes on this PR (Windows/macOS native-binary start steps can
      only be truly exercised by the real runners)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,26 +9,50 @@ on:
 jobs:
   build:
 
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-latest
     strategy:
       matrix:
-        os: [ ubuntu-latest, windows-latest, macOS-latest ]
-        python-version: [ '3.12' ]
+        python-version: [ '3.10', '3.11', '3.12', '3.13', '3.14' ]
         include:
-          - os: ubuntu-latest
-            python-version: '3.10'
-          - os: ubuntu-latest
-            python-version: '3.11'
-          - os: ubuntu-latest
-            python-version: '3.13'
-          - os: ubuntu-latest
-            python-version: '3.14'
-          - os: ubuntu-latest
-            python-version: '3.14'
+          - python-version: '3.14'
             minio: '7.2.18'
+
+    # Self-hosted MinIO instead of the public play.min.io playground,
+    # which has no uptime guarantee and has caused multi-hour CI hangs.
+    # Started as a plain `docker run` step below rather than a GitHub
+    # Actions `services:` container: `services:` has no way to pass a
+    # command, and this image has no default CMD, so it needs `server
+    # /data` passed explicitly. `services:` (Docker in general) is also
+    # only supported on Linux runners, so Windows/macOS (build-other-os
+    # below) skip the minio tests instead.
+    env:
+      AUDBACKEND_TEST_MINIO_HOST: localhost:9000
+      MINIO_ACCESS_KEY: minioadmin
+      MINIO_SECRET_KEY: minioadmin
+      MINIO_CONFIG_FILE: ${{ github.workspace }}/.github/minio-test.cfg
 
     steps:
     - uses: actions/checkout@v4
+
+    - name: Start MinIO
+      run: |
+        docker run -d --name minio -p 9000:9000 \
+          -e MINIO_ROOT_USER=minioadmin \
+          -e MINIO_ROOT_PASSWORD=minioadmin \
+          minio/minio:RELEASE.2025-09-07T16-13-09Z server /data --console-address ":9001"
+
+    - name: Wait for MinIO
+      run: |
+        for i in $(seq 1 30); do
+          if curl -sf http://localhost:9000/minio/health/live; then
+            echo "MinIO is up"
+            exit 0
+          fi
+          sleep 1
+        done
+        echo "MinIO did not become healthy in time"
+        docker logs minio
+        exit 1
 
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v5
@@ -47,7 +71,7 @@ jobs:
 
     - name: Test file-system only package
       run: |
-        uv run --no-default-groups --group test-filesystem --isolated pytest --cov-fail-under=0 tests/test_minimal_install.py tests/test_backend_filesystem.py 
+        uv run --no-default-groups --group test-filesystem --isolated pytest --cov-fail-under=0 tests/test_minimal_install.py tests/test_backend_filesystem.py
 
     - name: Test with pytest
       run: |
@@ -58,4 +82,40 @@ jobs:
       with:
         token: ${{ secrets.CODECOV_TOKEN }}
         file: ./coverage.xml
-      if: ${{ (matrix.os == 'ubuntu-latest') && (matrix.python-version != '3.14') }}
+      if: matrix.python-version != '3.14'
+
+  build-other-os:
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ windows-latest, macOS-latest ]
+        python-version: [ '3.12' ]
+
+    # No self-hosted MinIO here: GitHub Actions service containers
+    # (Docker) only work on Linux runners. Minio backend tests are
+    # skipped on these OSes rather than depending on play.min.io.
+    env:
+      AUDBACKEND_SKIP_MINIO_TESTS: '1'
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v5
+      with:
+        python-version: ${{ matrix.python-version }}
+
+    - name: Install uv
+      uses: astral-sh/setup-uv@3259c6206f993105e3a61b142c2d97bf4b9ef83d
+
+    - name: Sync Python environment
+      run: uv sync
+
+    - name: Test file-system only package
+      run: |
+        uv run --no-default-groups --group test-filesystem --isolated pytest --cov-fail-under=0 tests/test_minimal_install.py tests/test_backend_filesystem.py
+
+    - name: Test with pytest
+      run: |
+        uv run pytest --cov-config=.coveragerc.${{ runner.os }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,19 +12,13 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [ '3.10', '3.11', '3.12', '3.13', '3.14' ]
+        python-version: [ '3.10', '3.11', '3.12', '3.13' ]
         include:
+          - python-version: '3.14'
           - python-version: '3.14'
             minio: '7.2.18'
 
-    # Self-hosted MinIO instead of the public play.min.io playground,
-    # which has no uptime guarantee and has caused multi-hour CI hangs.
-    # Started as a plain `docker run` step below rather than a GitHub
-    # Actions `services:` container: `services:` has no way to pass a
-    # command, and this image has no default CMD, so it needs `server
-    # /data` passed explicitly. `services:` (Docker in general) is also
-    # only supported on Linux runners, so Windows/macOS (build-other-os
-    # below) skip the minio tests instead.
+    # Self-hosted MinIO instead of the public play.min.io playground.
     env:
       AUDBACKEND_TEST_MINIO_HOST: localhost:9000
       MINIO_ACCESS_KEY: minioadmin
@@ -93,8 +87,7 @@ jobs:
         python-version: [ '3.12' ]
 
     # Docker is not pre-installed on GitHub's Windows/macOS runners,
-    # so instead of a container these jobs run the native MinIO server
-    # binary -- same pinned release as the Linux job's Docker image.
+    # so we run the native MinIO server binary.
     # 127.0.0.1 is used instead of localhost throughout: on Windows,
     # localhost can resolve to IPv6 ::1 while the server listens on
     # IPv4 only, making every connection attempt fail.
@@ -124,13 +117,10 @@ jobs:
         $env:MINIO_ROOT_USER = "minioadmin"
         $env:MINIO_ROOT_PASSWORD = "minioadmin"
         New-Item -ItemType Directory -Force -Path "$env:RUNNER_TEMP\minio-data" | Out-Null
-        # Launch via a wrapper pwsh in its own (hidden) console. Using
-        # Start-Process with -RedirectStandardOutput directly makes the
-        # server inherit this step's console, and Windows terminates
-        # console-attached processes when the step's console closes --
-        # the server died at the step boundary. The wrapper's own
-        # console keeps it alive across steps; the log redirect happens
-        # inside the wrapper.
+        # Launch via pwsh in its own console.
+        # Using Start-Process with -RedirectStandardOutput
+        # makes the server inherit the console,
+        # and terminates console-attached processes at console closure.
         Start-Process -WindowStyle Hidden -FilePath pwsh -ArgumentList @(
           '-NoProfile', '-Command',
           "./minio-server.exe server `"$env:RUNNER_TEMP\minio-data`" --address 127.0.0.1:9000 *> minio.log"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,7 +43,7 @@ jobs:
 
     - name: Wait for MinIO
       run: |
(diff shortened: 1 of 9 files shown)


tests/test_backend_minio.py
```diff
@@ -285,6 +291,35 @@ def test_errors(host, repository, authentication):
...
+    os.environ["MINIO_CONFIG_FILE"] = config_path
+
+    with open(config_path, "w") as fp:
+        fp.write(f"[{host}]\n")
+        fp.write("access_key = user\n")
+        fp.write("secret_key = pass\n")
+        fp.write("secure = False\n")
+
+    backend = audbackend.backend.Minio(host, "repository")
+    assert backend._client._base_url.is_https is False
```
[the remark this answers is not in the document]
ChristianGeng: Addressed all three:

1. `MINIO_CONFIG_FILE` leak: not actually an issue here — `hide_credentials` already saves/restores it in its teardown (same pattern every other test in this file that sets `MINIO_CONFIG_FILE` directly relies on, e.g. `test_authentication`, `test_get_config`), so it's cleaned up regardless. Did adopt the parametrize idea though — expanded into `test_secure_config_string_values`, covering the full set of false-like (`0`, `no`, `off`, `""`) and true-like (`1`, `yes`, `on`) strings the actual parsing logic checks, not just `"False"` alone. Good catch on the test being too narrow.
2. Pragma syntax: already caught and fixed in this same PR (see the commit message on 6f43c8e) — coverage.py's exclude patterns match the literal string `pragma: no <OS> cover`, so a single `pragma:` prefix shared across both OS names doesn't match either. Now `# pragma: no Windows cover  pragma: no macOS cover`, verified against real CI on both platforms.
3. Bumped the wait loop from 30 to 60 attempts.

.github/workflows/test.yml
```diff
@@ -9,27 +9,51 @@ on:
...
-          - os: ubuntu-latest
-            python-version: '3.11'
-          - os: ubuntu-latest
-            python-version: '3.13'
-          - os: ubuntu-latest
-            python-version: '3.14'
-          - os: ubuntu-latest
-            python-version: '3.14'
+          - python-version: '3.14'
             minio: '7.2.18'
```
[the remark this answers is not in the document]
ChristianGeng: Not a duplicate by accident — this is pre-existing, intentional test design, not something introduced by this PR. Checked `main`'s test.yml before this change: the same two separate `3.14` entries (one plain, one with `minio: '7.2.18'`) were already there. The intent is to test the newest Python against both the current minio-py client and an older pinned one (`7.2.18`), on separate runs. Leaving as-is.

.github/workflows/test.yml
```diff
@@ -9,27 +9,51 @@ on:
...
 
+    # Self-hosted MinIO instead of the public play.min.io playground,
+    # which has no uptime guarantee and has caused multi-hour CI hangs.
+    # Started as a plain `docker run` step below rather than a GitHub
+    # Actions `services:` container: `services:` has no way to pass a
+    # command, and this image has no default CMD, so it needs `server
+    # /data` passed explicitly. `services:` (Docker in general) is also
+    # only supported on Linux runners, so Windows/macOS (build-other-os
+    # below) skip the minio tests instead.
```
hagenw: This reads as if we would do no tests under Windows and MacOS on MinIO, but below we start MinIO for both as well.

```suggestion
    # Self-hosted MinIO instead of the public play.min.io playground.
```

.github/workflows/test.yml
```diff
@@ -58,4 +82,96 @@ jobs:
...
+      matrix:
+        os: [ windows-latest, macOS-latest ]
+        python-version: [ '3.12' ]
+
+    # Docker is not pre-installed on GitHub's Windows/macOS runners,
+    # so instead of a container these jobs run the native MinIO server
+    # binary -- same pinned release as the Linux job's Docker image.
+    # 127.0.0.1 is used instead of localhost throughout: on Windows,
+    # localhost can resolve to IPv6 ::1 while the server listens on
+    # IPv4 only, making every connection attempt fail.
```
hagenw, line 93:
```suggestion
    # Docker is not pre-installed on GitHub's Windows/macOS runners,
    # so we run the native MinIO server binary.
    # 127.0.0.1 is used instead of localhost throughout: on Windows,
    # localhost can resolve to IPv6 ::1 while the server listens on
    # IPv4 only, making every connection attempt fail.
```

.github/workflows/test.yml
```diff
@@ -58,4 +82,96 @@ jobs:
...
+        $env:MINIO_ROOT_USER = "minioadmin"
+        $env:MINIO_ROOT_PASSWORD = "minioadmin"
+        New-Item -ItemType Directory -Force -Path "$env:RUNNER_TEMP\minio-data" | Out-Null
+        # Launch via a wrapper pwsh in its own (hidden) console. Using
+        # Start-Process with -RedirectStandardOutput directly makes the
+        # server inherit this step's console, and Windows terminates
+        # console-attached processes when the step's console closes --
+        # the server died at the step boundary. The wrapper's own
+        # console keeps it alive across steps; the log redirect happens
+        # inside the wrapper.
```
hagenw:
```suggestion
        # Launch via pwsh in its own console.
        # Using Start-Process with -RedirectStandardOutput
        # makes the server inherit the console,
        # and terminates console-attached processes at console closure.
```

audbackend/core/backend/minio.py
```diff
@@ -57,13 +57,17 @@ class Minio(Base):
     .. _minio.Minio: https://min.io/docs/minio/linux/developers/python/API.html
 
     Examples:
+        The network calls below are not run as part of the test suite
+        (``# doctest: +SKIP``), since they depend on the public
+        play.min.io playground being reachable, which CI cannot rely on.
+
```
hagenw: The `# doctest: +SKIP` is not visible in the user documentation, so we should also not add this comment here.


```suggestion
```

audbackend/core/backend/minio.py
```diff
@@ -90,6 +94,11 @@ def __init__(
         config = self.get_config(host)
         if secure is None:
             secure = config.get("secure", True)
+            if isinstance(secure, str):
+                # ``configparser`` returns config file values as strings,
+                # e.g. ``"False"``, which is truthy in Python
+                # and would otherwise silently enable HTTPS.
+                secure = secure.strip().lower() not in ("false", "0", "no", "off", "")
```
hagenw, line 97: We used the `secure` setting before when testing a MinIO server on our own infrastructure, why didn't we had the problem there?

But I guess it is not too important to solve this here, as the problem will anyway go away when we switch to use `audeer.load_configuration()` in some follow up pull request.

tests/conftest.py
```diff
@@ -14,11 +14,22 @@
...
 pytest.UID = audeer.uid()[:8]
 
-# Define static hosts
+# Define static hosts.
+#
+# ``AUDBACKEND_TEST_MINIO_HOST`` lets CI point the minio backend tests
+# at a self-hosted MinIO instance (e.g. a GitHub Actions service
+# container) instead of the public play.min.io playground, which has
+# no uptime guarantee. Defaults to play.min.io for local development,
+# where standing up a MinIO server isn't expected.
```
hagenw:
```suggestion
# Use play.min.io when running tests locally,
# and point to ``AUDBACKEND_TEST_MINIO_HOST`` in CI
# (a self-hosted MinIO instance in GitHub Actions).
```

tests/conftest.py
```diff
@@ -14,11 +14,22 @@
...
 pytest.HOSTS = {
-    "minio": "play.min.io",
+    "minio": os.environ.get("AUDBACKEND_TEST_MINIO_HOST", "play.min.io"),
 }
 
+# Skip minio backend tests entirely where no MinIO server is available.
+# GitHub Actions service containers (Docker) only work on Linux runners,
+# so Windows/macOS CI jobs set this rather than depending on play.min.io.
+pytest.SKIP_MINIO = os.environ.get("AUDBACKEND_SKIP_MINIO_TESTS") == "1"
+
```
hagenw: `AUDBACKEND_SKIP_MINIO_TESTS` is not set in `.github/workflows/test.yml` and we run tests under all platforms. So, I guess you can remove `pytest.SKIP_MINIO` here and from the tests?

.github/workflows/test.yml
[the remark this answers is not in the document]
hagenw: There is a difference to how it was before: Python 3.14 is now only tested for MinIO 7.2.18, but never for MinIO's latest version. Before we had two jobs for Python 3.14, now we have just one.

tests/test_backend_minio.py
```diff
@@ -285,6 +291,57 @@ def test_errors(host, repository, authentication):
...
+        ("false", False),
+        ("0", False),
+        ("no", False),
+        ("off", False),
+        ("", False),
+        ("True", True),
+        ("true", True),
+        ("1", True),
+        ("yes", True),
+        ("on", True),
```
hagenw, line 301: Can you add tests for an empty `secure:` line and a typo:


```suggestion
        ("False", False),
        ("false", False),
        ("0", False),
        ("no", False),
        ("off", False),
        ("", False),
        ("True", True),
        ("true", True),
        ("1", True),
        ("yes", True),
        ("on", True),
        (None, False),
        ("typo", False),
```

hagenw (commented):
Cool, was not aware that you can run your own MinIO inside Github Actions.

What is untested now in CI is the https route, but I think this is ok as on the positive side we no longer depend on play.min.io, which can indeed be unavailable for some time.

tests/conftest.py
hagenw: Maybe we should also document here that host + credentials + secure flag need to be set when running the tests not on play.min.io, e.g. a developer could try to run them locally with:

```
AUDBACKEND_TEST_MINIO_HOST=localhost:9000 uv run pytest tests/test_backend_minio.py
```

Maybe we should first add configuration handling with `audeer.load_configuration()` then you can also set `secure` via an environment variable and don't have to create the akward config file?

hagenw: We are not using YAML files here and have anyway another structure for config handling, so using `audeer.load_configuration()` makes no sense.

Maybe just document that you need to setup credentials and a config file. Or we simply stay with the current approach as it is very unlikely that a developer will try her/his own local MinIO instance when running the tests.

.github/workflows/test.yml
ChristianGeng: Applied in 339321b.

.github/workflows/test.yml
ChristianGeng: Applied in 339321b.

.github/workflows/test.yml
ChristianGeng: Applied in 339321b.

audbackend/core/backend/minio.py
ChristianGeng: Applied in 339321b; agree the test-setup note does not belong in user-facing docs.

tests/conftest.py
ChristianGeng: Applied in 339321b.

tests/conftest.py
ChristianGeng: Agreed, it was dead code once the minio tests started running on every platform.
Removed `pytest.SKIP_MINIO` and its uses in 339321b.

.github/workflows/test.yml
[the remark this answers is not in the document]
ChristianGeng: You are right, and it was a regression introduced by this PR: an `include:` entry whose keys match an existing matrix combination is merged into that combination instead of creating a new job, so `{python-version: '3.14', minio: '7.2.18'}` only added the pin to the already existing 3.14 job.
Fixed in 339321b by moving 3.14 out of the base list into two explicit `include:` entries, so 3.14 is again tested both against the current minio client and against the pinned 7.2.18.

audbackend/core/backend/minio.py
ChristianGeng: Only the false-like case was broken. `configparser` returns strings, so `secure = True` arrived as the string `"True"`, which is truthy and therefore behaved correctly by accident — any true-like value worked.
The bug was only observable when someone actually wanted to turn TLS off, so testing an own MinIO instance over HTTPS would never have surfaced it.

tests/conftest.py
ChristianGeng: Agreed, let's stay with the current approach for this PR.
The conftest comment now names `AUDBACKEND_TEST_MINIO_HOST`, so a developer pointing the tests at a local instance has the entry point.

tests/test_backend_minio.py
ChristianGeng: Both suggested cases fail against the current code, so they are a behaviour change rather than extra coverage.
The test body writes `f"secure = {secure_value}"`, so `None` ends up in the config as the literal string `"None"`, which is not in the false-like set and yields `is_https=True`; `"typo"` yields `is_https=True` for the same reason, since the current rule is "recognised false-like value means HTTP, anything else means HTTPS".

Treating an unrecognised value as `False` would silently downgrade the connection to plain HTTP, which is the less safe direction for a typo. The alternative is to follow the `_parse_timeout` precedent in the same module: warn on an unrecognised value and fall back to the secure default, covered by a test like `test_invalid_timeout_warning`. Which do you prefer?

Independent of that decision, `secure =` (empty value) and a bare `secure` with no value already behave as `False` today, so those two cases are coverable immediately either way.

audbackend/core/backend/minio.py
hagenw: But for our internal MinIO server we needed `secure = False` that was the main reason we introduced `secure` before?

tests/test_backend_minio.py
hagenw: Ok, I thought `None` would be the case `secure =` (empty value). Is this case covered in any test?

With the typing I agree, we can stay with the current behavior, but then adding `("typo", True)` to the test should work and document that behavior.

ChristianGeng: Yes, this is covered: 

The `("", False)` entry writes `secure =` with an empty value, because the test builds the line with `f"secure = {secure_value}"`.

A bare `secure` with no value is the same case: `configparser` (created with `allow_no_value=True`) turns both into the empty string, so they reach the parsing code as the same input. 

So a separate test would just do the same thing.

Added `("typo", True)` in 35ac05b though.

hagenw (approved):

audbackend/core/backend/minio.py
ChristianGeng: The `secure` **argument** always worked; only reading `secure` from the **config file** was broken.
The bug had exactly one trigger: a false-like value in a config file.

So it works - but not from the config file.

